Issue #4, opened 2026-08-20
Phase 5 — Release gate and first release


onemen:
Phase 5 — Release gate and first release

The v1.0 release gate covers production publishing, user-facing documentation, signing, and final CI validation.

Done:
- [x] First prod publish → installer + utils/fx-folder/updater-ui zips on the release page (via pages.yml)
- [x] Release page points the community at `latest`
- [x] GitHub environment protection: `upload --mode=prod` runs only after the E2E workflow passes
- [x] `upload --mode=prod` moves the `latest` tag to the uploaded commit (P0-1)
- [x] Run the cache-prune script as part of the release gate
- [x] Release versioning decided — ADR 0019

Still open:
- [x] Persistent footer in the installer UI (`installer/web/index.html`) and the updater tab (`tools/publish/remote-ui/updater.html`) — always visible, with a copyright/license line (mind the MPL-2.0 core vs MIT updater split) and links to the issue tracker and the docs (the repo README today). The updater tab is XML-parsed XHTML (no `innerHTML`), so the footer is static markup + CSS in `remote-ui`; the installer footer ships via the embedded `installer/web` assets. Users get it through the installer binaries and `updater-ui.zip`.
- [ ] Pre-1.0 release candidate (RC) of the installer + updater — publish an RC build so the community can validate the real install/self-update flow before `latest` points at v1.0. Mechanism decided (ADR 0026, merged #189): `--mode=dev` stays branch-only; `--tag` + `--note` on a dev publish creates a named RC release on the dev channel; `--note` alone just labels the branch/page. RC artifacts + manual checks (installer E2E all OSes, updater self-update) go on the release checklist.
- [x] Browser-version drift gate trusts only a validated-versions record populated by successful browser-specific E2E — implemented: e2e.yml `record-validation` job writes .watchdog-validated, pages.yml prod pre-flight blocks on drift (PR #92)
- [ ] Decide: zip the installer executable, or provide a separate README.txt release asset? Do not treat zip packaging as a SmartScreen/UAC workaround; MOTW can propagate through extracted files.
- [ ] Research free/cheap code-signing options and document the CI integration plan:
  - **SignPath OSS program** — free Windows Authenticode signing for eligible open-source projects; CI would upload the built installer to SignPath, wait for signing, then download the signed artifact into the release upload.
  - **Azure Trusted Signing** — Microsoft's cloud signing service; CI would authenticate with an Azure service identity, submit the installer, wait for signing, then publish the signed artifact.
[...]

PR #40, opened 2026-08-24
feat(publish): move the latest tag to the uploaded commit on prod upload

onemen:
Part of #4 (release gate).

After every non-idle prod upload, force-update the `latest` release tag to the uploaded commit
(HEAD) and print old → new SHA; `--no-tag` is the escape hatch. The installer/updater fetch via
`.../releases/download/latest`, so the tag name stays stable while its target follows each
publish. Idle runs and dev mode are untouched.

### DIFF
--- a/tools/publish/upload.mjs
+++ b/tools/publish/upload.mjs
@@ -26,6 +26,9 @@
 //                      detached worktree (your checkout is left untouched).
 //   --ci               build binaries for all platforms (default: current OS).
 //   --platform=win|linux|mac (repeatable)  explicit binary platform set.
+//   --no-tag           (prod only) skip moving the 'latest' release tag to the
+//                      uploaded commit (it is force-updated after every
+//                      non-idle prod upload).
 //   --verbose          per-file zip listings and other detail lines.
 //   --quiet            suppress progress output (errors still print).
 //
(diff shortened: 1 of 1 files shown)


tools/publish/upload.mjs
```diff
@@ -514,6 +521,32 @@ async function publishToGitHub({
...
     }
   }
+
+  // Prod: keep the 'latest' release tag pointing at the commit this upload was
+  // built from (the current HEAD — prod is main-only with a clean worktree).
+  // The installer/updater fetch everything via
+  // .../releases/download/latest, so the tag NAME must stay stable while its
+  // target follows each publish; an idle run (nothing rebuilt) leaves it where
+  // it is, since the release assets did not change either. --no-tag skips.
+  if (PUBLISH_MODE === 'prod' && release && !NO_TAG && anythingUploaded) {
```
onemen, line 531: **CodeRabbit (combined local review) — major**

Update the tag-reference flow around octokit.git.getRef and octokit.git.updateRef so a 404 creates the missing tag reference using the GitHub create-reference operation, while existing references continue through updateRef. Preserve rethrowing non-404 errors and use headSha as the target for both paths.

onemen: Fixed: `updateRef` now calls `octokit.git.createRef` when `getRef` returns 404, so the `latest` tag is created on first use.

Commit: 61ff6cd

PR #41, opened 2026-08-24
feat(publish): add dev-clean for removing dev-build branches and tags

onemen:
P0-2 on the v1.0 roadmap.

`pnpm dev-clean --id <id>` deletes one dev build, `--all` prunes every `dev-build-*` branch +
tag, locally and on origin (GitHub API), with `--dry-run` / `--local-only` / `--remote-only`
scopes. Unknown ids fail loudly; the current branch is never deleted. Pure arg/selection logic is
unit-tested (`tools/test/unit/devClean.test.mjs`).

### DIFF
--- a/tools/publish/devClean.mjs
+++ b/tools/publish/devClean.mjs
@@ -128,10 +128,14 @@ export async function listRemoteRefs(octokit) {
  * }}
  */
 export function selectTargets(local, remote, opts) {
+  // The lists arrive pre-filtered (listLocalRefs/listRemoteRefs), but re-check
+  // the prefix so a non-dev-build ref can never be selected even if a future
+  // caller stops pre-filtering.
   const pick = list =>
-    opts.all ? [...list]
+    (opts.all ? [...list]
     : list.includes(opts.id) ? [opts.id]
-    : [];
+    : []
+    ).filter(n => n.startsWith(DEV_PREFIX));
   return {
     local: {branches: pick(local.branches), tags: pick(local.tags)},
     remote: {branches: pick(remote.branches), tags: pick(remote.tags)},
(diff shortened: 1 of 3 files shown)


tools/test/unit/devClean.test.mjs
```diff
@@ -0,0 +1,71 @@
...
+  assert.deepEqual(hit.local.branches, ['dev-build-main-aaa']);
+  assert.deepEqual(hit.remote.tags, ['dev-build-main-aaa']);
+  assert.deepEqual(hit.remote.branches, []);
+
+  const miss = selectTargets(local, remote, {id: 'dev-build-main-zzz', all: false});
+  assert.deepEqual(miss.local.branches, []);
+  assert.deepEqual(miss.remote.tags, []);
+});
+
+test('selectTargets: --all selects every dev-build ref', () => {
```
onemen: **CodeRabbit (combined local review) — minor**

Update the selectTargets --all test to include non-dev-build branches and tags in both local and remote fixtures, then assert those refs are excluded while all dev-build refs remain selected.

tools/publish/devClean.mjs
```diff
@@ -0,0 +1,268 @@
...
+        `No dev build matches '${opts.id}'. ` +
+          (existing.length ?
+            `Existing dev-build refs:\n  ${existing.join('\n  ')}`
+          : 'No dev-build refs exist.')
+      );
+    }
+  }
+
+  let ok = true;
+  for (const name of targets.local.branches) ok = deleteLocalBranch(name, opts.dryRun) && ok;
```
onemen, line 251: **CodeRabbit (combined local review) — minor**

Update the remote branch deletion flow around deleteRemoteRef so the repository’s current branch is excluded from targets.remote.branches, including when --remote-only is used with that branch ID; retain deletion of other remote branches and remote tags.

onemen: Fixed: `selectTargets` now excludes the current branch from remote targets (matching the existing local-branch guard).

Commit: 6458461

tools/test/unit/devClean.test.mjs
onemen: Fixed: the `--all` test now includes non-dev-branch fixtures and asserts they are excluded from targets.

Commit: 6458461

PR #43, opened 2026-08-24
fix(core): toggle boolean attributes in createElement on Firefox 149+

onemen:
P2-1 on the v1.0 roadmap (core fix, reviewed last).

Firefox 149+ (bug 2008041) evaluates boolean attributes by presence rather than value, so
`checked="false"` is still treated as checked. On 149+ route boolean-like attribute values through
`toggleAttribute` (absent = off); older builds keep the value-based `setAttribute` behavior.
Reported upstream in xiaoxiaoflood/firefox-scripts#405.

## Testing

Mock-DOM unit tests (`tools/test/unit/attributeUtils.test.mjs`) cover the boolean-attribute semantics directly: a values × ff149 matrix (`true` / `false` / `"true"` / `"false"` / other values → the exact `setAttribute`/`toggleAttribute` call) and the version gate with mocked appinfo — no real Firefox needed. The gate reads `appinfo.platformVersion` (the Gecko version) so forks with their own release numbering (LibreWolf, Waterfox, Floorp, Zen) still follow the Gecko behavior.


core/chrome/utils/userChrome.js
```diff
@@ -16,6 +16,18 @@ const UC = {
   sandboxes: new WeakMap(),
 };
 
+// Firefox 149+ (bug 2008041) evaluates boolean attributes by PRESENCE rather
```
onemen: **CodeRabbit (combined local review) — major**

Add a targeted regression test for the Firefox 149 behavior in createElement, covering boolean attributes and "true"/"false" values under both pre-149 and 149+ version paths; if this cannot be tested in the existing framework, document the limitation in the PR description.

onemen: Fixed: added `tools/test/unit/userChrome.test.mjs` — evaluates the full `userChrome.js` in a Node vm with mocked Firefox globals, covering the FF149 boolean-attribute semantics, fork `platformVersion` detection, and `createElement` attribute routing.

Commit: c6c30e0

PR #44, opened 2026-08-24
perf(e2e): centralize browser downloads and shorten updater waits

onemen:
## What & why

P0-3 from the v1.0 roadmap — make the E2E suite fast enough for the release gate and for everyday PR runs.

- **`tools/test/e2e/downloads.mjs` (new):** one data file mapping each browser → per-OS install method using official download URLs / package managers (Mozilla tarball, brew cask, choco, official pages for Waterfox/Zen/Dev Edition), instead of hardcoding install steps in the workflow. Includes a `--list`/CLI and `installBrowser()` returning the resolved binary path.
- **`.github/workflows/e2e.yml`:** the three per-OS Firefox install blocks are replaced by a loop over the downloads map.
- **Shorter waits** in `updater-e2e.mjs` / `helpers.mjs`: tab-open deadline 90s → 15s, no-tab window 30s → 10s, card render 60s → 15s. The scheduler runs at startup, so a tab that hasn't opened in ~15s won't — the fast paths (config-probe mirror line, `lastUpdateTabShown`) already return in ~1–5s on success.

## CodeRabbit (round 1, batch review)

Finding posted for tracking; the fix is included in this branch:

- `fetchWithRetry` was unbounded — a stalled download would hang CI until the runner killed the job. Each attempt now uses `AbortSignal.timeout` (5 min) and flows through the existing retry/logging path (`8479b45`).

### DIFF
--- a/tools/test/e2e/downloads.mjs
+++ b/tools/test/e2e/downloads.mjs
@@ -1,0 +1,218 @@
+#!/usr/bin/env node
+
+/**
+ * tools/test/e2e/downloads.mjs — official browser download/install map for the
+ * E2E CI matrix.
+ *
+ * The workflow (e2e.yml) used to hard-code each browser's install commands per
+ * OS; this module is the single source of truth: browser → per-OS install
+ * recipe, plus a CLI that performs the install for the current OS and exports
+ * the resolved binary path.
+ *
+ * CLI (used by e2e.yml, works locally too): node tools/test/e2e/downloads.mjs
+ * <browser> [--os win|mac|linux]
+ *
+ * On success the resolved binary path is printed to stdout and, when running
+ * inside GitHub Actions ($GITHUB_ENV set), appended to $GITHUB_ENV as
+ * FIREFOX_BINARY — the updater E2E (updater-e2e.mjs) reads that env var.
+ *
+ * Browsers without an automated recipe (manual install only) list their
+ * official download page instead; the CLI fails with a clear message.
+ */
+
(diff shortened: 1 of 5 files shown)


tools/test/e2e/downloads.mjs
```diff
@@ -0,0 +1,221 @@
...
+  // Extract next to existing content (tar xf, no strip): $HOME/firefox-app/firefox/
+  execSync(`tar xf "${archive}" -C "${dest}"`, {stdio: 'inherit'});
+  const binary = path.join(dest, 'firefox', 'firefox');
+  if (!fs.existsSync(binary)) {
+    throw new Error(`tarball extracted, but ${binary} not found`);
+  }
+  return binary;
+}
+
+async function fetchWithRetry(url, attempts, timeoutMs = 300_000) {
```
onemen, line 130: **CodeRabbit (combined local review) — major**

Update `fetchWithRetry` to bound each fetch attempt with a timeout by passing an `AbortSignal.timeout` value to fetch. Keep timeout failures flowing through the existing catch, logging, delay, and retry behavior, while preserving successful responses and HTTP error handling.

_Addressed in `8479b45` — each attempt now aborts after 5 minutes via `AbortSignal.timeout`._

onemen: Fixed: `downloads.mjs` now passes `AbortSignal.timeout(5 * 60_000)` to every fetch attempt, so no download can hang CI indefinitely.

Commit: 8479b45

PR #67, opened 2026-08-26
ci: guard the publish gate always-report design in ci-gate

onemen:
Mirror of the e2e-gate browser-matrix assertion from #66.

`ci-gate` now verifies, in the no-publish-relevant branch, that the publish gate job reported **success** — the always-run no-op it's designed for. If the gate ever gains a job-level `if:` (a plausible 'cleanup'), it would skip on docs-only PRs and make the required check go missing, blocking strict branch protection — the exact #65 regression. The guard turns that silent regression into a loud CI failure.

Shell logic verified locally for all five cases (legit ×2, regression ×3).


onemen:

PR #68, opened 2026-08-26
ci: extend the e2e-gate skip guard to installer + updater

onemen:
Extends the runtime gate-removal guard from #66 from browser-matrix to **all three** gated E2E jobs (installer, updater, browser-matrix), one loop that names the regressed job.

Shell logic verified locally for all 8 gate cases: legit ×2 exit 0; each removed gate ×3 exit 1; installer/updater failure exit 1; browser-matrix failure stays advisory (warning, exit 0).


onemen:

PR #69, opened 2026-08-26
ci: log the URL watchdog baseline's cache-hit status and age

onemen:
The watchdog baseline lives in the Actions cache, which evicts/expires silently (7-day TTL). An eviction previously looked identical to a first run or a version bump — the real cause was invisible in the log.

Each run now prints one line up front:
- `baseline: cache hit — written <ISO ts> (<age> ago)` (age from the restored file's mtime)
- `baseline: cache miss — first run, re-baselining all browsers`

`formatAge()` is exported + unit-tested (4 cases incl. negative clamp); both log lines smoke-verified live. Docs (DEVELOPING.md) updated.


onemen:

PR #70, opened 2026-08-26
ci: consolidate gate guards into one shared action + static contract

onemen:
One PR for the whole CI-hardening batch — supersedes #67, #68 and #69 (closed). Since every change touches workflow files, the path filter runs the full matrix on the PR itself, so the new gate wiring validates itself live.

## What's in it

**Shared gate engine** — `.github/actions/verify-gate` (composite action + `verify.sh`) replaces the duplicated inline shell in `ci-gate` / `e2e-gate`: required / advisory / skip-guard / always-report checks. The skip-guard now covers **every** gated E2E job (a removed path-filter `if:` fails the next docs-only PR, naming the job); the always-report guard preserves the publish-gate design (a reintroduced job-level `if:` fails instead of re-breaking strict branch protection).

**Full path filtering** — snapshot (dev-snapshot build) and helper (elevated-copy) joined installer/updater/browser-matrix: docs-only PRs now skip **all** E2E jobs, not just the big ones.

**Static contract** — `pnpm check:gates` (wired into the `checks` job) asserts: every workflow job is in its gate's `needs:`, path-filter `if:`s are in place, always-report jobs carry no job-level `if:`, gates carry `if: always()`. A new job can no longer silently bypass the gate. 6 unit tests pin the parser + rules.

**Merge queue readiness** — `merge_group` triggers on ci/e2e/url-watchdog so required checks run on the queue's temporary branch. Enabling the queue in repo settings (Settings → General → merge queue; branch protection requiring it) lands PRs by rebasing heads onto main, sidestepping the update-branch over-trigger documented in ADR 0017.

**Watchdog telemetry** — each weekly run logs `baseline: cache hit — written <ts> (<age> ago)` or `baseline: cache miss — first run`, so a silently evicted baseline cache is visible.

ADR 0017 + DEVELOPING.md updated to match.

## Validation

- `verify.sh` engine simulated live across 8 scenarios (legit ×2 per gate, advisory, each guard firing, changes-failure) — plus newline-separated results from the literal-block `results:` input.
- `pnpm test`: 114 pass / 0 fail; `pnpm check:gates` ✓; `pnpm check:decisions` ✓; `pnpm format` ✓; `pnpm lint` ✓.

## Settings to flip (after merge)

1. Settings → General → **merge queue**: enabled.
2. Branch protection: require the merge queue; keep the required checks as-is (CI gate, E2E gate, … — they already report on `merge_group`).

## Summary by CodeRabbit

* **New Features**
  * CI, E2E, and URL monitoring now support merge-queue validation.
  * E2E checks are selectively skipped for documentation- and tooling-only changes.
  * Added automated validation for workflow gate coverage and reporting.
  * Browser-download checks now display cache age and status.

* **Bug Fixes**
  * Improved handling of missing or outdated browser-download baselines.
  * Corrected human-readable duration formatting, including clock-skew values.

* **Documentation**
  * Updated CI guidance and validation decision records for merge queues, path filtering, and advisory checks.

### DIFF
--- a/.github/actions/verify-gate/action.yml
+++ b/.github/actions/verify-gate/action.yml
@@ -52,4 +52,6 @@ runs:
         ADVISORY: ${{ inputs.advisory }}
         SKIP_GUARD: ${{ inputs.skip-guard }}
         ALWAYS_REPORT: ${{ inputs.always-report }}
-      run: ${{ github.action_path }}/verify.sh
+      # `bash <script>` (not a bare path): the exec bit is not stored for
+      # files added from Windows, so a direct run would get Permission denied.
+      run: bash ${{ github.action_path }}/verify.sh
--- a/.github/actions/verify-gate/action.yml
+++ b/.github/actions/verify-gate/action.yml
@@ -1,0 +1,50 @@
+name: Verify gate
+description: >-
+  Shared engine for the aggregate CI/E2E branch-protection checks. Verifies the changes
+  (paths-filter) job, then the branch-conditional gated jobs (required / advisory), then the
+  no-change guards (skipped-jobs and always-report). This is the shell logic that used to live
+  inline in ci-gate and e2e-gate.
+inputs:
+  changes-result:
+    description: result of the changes (paths-filter) job — must be success
+    required: true
+  branch:
+    description: "'true' when gated work is active (the filter output), else 'false'"
+    required: true
+  branch-label:
+    description: human label for the branch, e.g. 'E2E-relevant changes'
+    required: true
+  results:
+    description: 'space-separated name:result pairs for every tracked job'
+    required: true
+  required:
+    description: 'space-separated job names that must be success when branch=true'
+    required: false
+    default: ''
+  advisory:
+    description: 'space-separated job names warned (not failed) when branch=true'
+    required: false
+    default: ''
+  skip-guard:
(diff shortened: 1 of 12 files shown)


onemen:
CodeRabbit findings status (reviewed in commit c3a1978; full re-review pending):

1. **Missing checkout before the local action** (Critical) — fixed: both gates run a pinned `actions/checkout` before `verify-gate`. Also fixed the follow-on `Permission denied` (exec bit not stored from Windows → `bash ${action_path}/verify.sh`).
2. **Gate wiring** (Major) — fixed: `check-gate-coverage` now parses the verify-gate `with:` block and asserts needs ↔ results ↔ classification coverage (+3 unit tests).
3. **Immutable cache key** (Major) — fixed: baseline keys rotate per run (`url-watchdog-baseline-${{ github.run_id }}` + `restore-keys` prefix), so the updated baseline actually persists.
4. **Merge-queue wording** (Minor) — fixed in DEVELOPING.md + ADR 0017 (merge-group validation, not head rebasing).

Also fixed a regression found while fixing #1: `checks` (always-run) is now verified in both branches via a new `always-verify` input — previously a failed lint job on a docs-only PR would have slipped through the gate.

All 22 CI checks pass; happy to re-run the review whenever convenient.

PR #106, opened 2026-09-04
fix(lint): make the analyzer gate actually run -fanalyzer (+ result cache)

onemen:
## Summary
**Stacked on #105** — base is `perf/lint-time`; do not merge before it. The diff is exactly the delta on top of that branch (GitHub shows it automatically against the stacked base).

**The gate never analyzed anything.** `-fsyntax-only` disables gcc's `-fanalyzer` pass entirely on modern gcc, so `make analyze` verified syntax only while reporting itself as a memory-safety gate. `tools/analyze-c.mjs` now really **compiles** each file (`-c`, objects in a temp dir), keeping the parallel per-file runner.

**Real findings surfaced and fixed** (`installer/src/`):
- `detect_browser.c` `json_value_after_key()`: wrote `out[0] = '\0'` **before** the NULL guard — a genuine dereference-before-check.
- Two `strncpy` INI parsers (`check_compatibility_ini`, `lookup_profile_by_name`): `strncpy` reads its full `n` range, over-reading past the 512-byte `line` buffer → replaced with strlen-bounded `memcpy`.
- `main.c` / `http_server.c`: two **documented** `#pragma GCC diagnostic` suppressions for one confirmed winsock false positive (the analyzer models a handle that is simultaneously created and `INVALID_SOCKET`; every real handle is closed). Each pragma carries a "verify by removing this and running make analyze" note.

**Result cache**: `installer/.analyzer-cache/` (gitignored) keys on the content of every transitively included header — *not* mtime, because the Makefile re-touches generated headers each run — plus source bytes, flags, gcc version. Only **clean** runs are cached (gcc exits 0 even with warnings); cached findings are replayed so a dirty tree still fails. Warm `make analyze` 3.4s → **0.5s**.

## Validation
- Gate green on Windows (UCRT64 gcc 16, 3.4s cold / 0.5s warm) **and** on the POSIX branches via WSL gcc 15.2 — which matters: CI's `checks` job runs `pnpm lint` on ubuntu, so the POSIX branches are what CI compiles.
- Negative tests: an injected `malloc` leak and a syntax error both fail the gate; the leak is never cached.
- `make dist_win` builds; 138 unit tests pass; eslint/prettier clean.
- `pnpm review:local`: 0 findings after the file-list fix (hard-coded SOURCES → derived from `src/*.c`).

Fixes the latent issue surfaced in #105's notes.

🤖 Generated with Codebuff


onemen (commented):
Superseded — finding re-posted as a line-anchored thread on this PR.

tools/analyze-c.mjs
```diff
@@ -67,63 +132,92 @@ function runGcc(file) {
...
-      // Hard compiler error (syntax error, bad flag, …) — always a failure,
-      // analyzer findings or not.
-      console.error(`✗ gcc failed on ${file} (exit ${code}):`);
-      console.error(output || '(no output)');
-      process.exit(1);
+    const unrunnable = results.find(r => r.code === null);
+    if (unrunnable) {
+      console.error('analyze-c failed: gcc could not be run.');
+      console.error(unrunnable.output);
+      process.exitCode = 2;
```
onemen, line 172: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit `review:batch` run)

**Minor — right.** `process.exit()` on the failure paths skipped the `try/finally` cleanup, so every failing `make analyze` run leaked an `analyzer-*` temp dir. Fixed in ade7804: the failure paths set `process.exitCode` and return, so the cleanup runs. Verified with an injected leak probe: the gate exits 2 and leaves zero temp dirs.

Issue #30, opened 2026-08-24
Core code test coverage (stub smoke tests + Nightly leg + test-required convention)


onemen:
core/** (userChrome.js, BootstrapLoader.js, config.js) has no automated coverage today.

- Level 1 (in `pnpm test`, PR gate): Node stub smoke tests — stub Services/ChromeUtils/IOUtils enough to load userChrome.js, BootstrapLoader.js and config.js top-level; assert no throw. Catches syntax + API drift cheaply.
- Level 2 (e2e harness): real-browser "core smoke" leg — boot stable + Nightly with installed utils; assert the userChromeJS loader ran and no startup errors; scheduled (nightly cron) to catch upstream Firefox changes.
- Level 2b — BootstrapLoader load verification: **delivered as a draft** on branch `postv1/manifest-lifecycle-e2e` (post-v1 deferral, same as the rest): `test/e2e/core/manifest-lifecycle-e2e.mjs`, run via `pnpm test:e2e:legacy`, **13/13 passing** on Firefox 155.0.1. It seeds a profile with the shipped utils **verbatim** (no source patching), installs a real legacy bootstrap extension, and asserts `chrome://testext/content/test.html` stays readable across three sessions — fresh profile, plain restart, restart after clearing `startupCache` + `cache2` (the "clear cache and restart" ritual). It also seeds stale `chrome.manifest*` crash litter and asserts the startup sweep removes it while the extension's chrome still comes up (the sweep itself was added to BootstrapLoader.js on the same branch; task tracked in #38). CI wiring stays deferred with the gate below.
- Then enable the mechanical "core changed && no test changed → fail" CI gate.

### Recommended follow-up tests (from the discussion #101 manifest-lifecycle investigation)

- **Delayed-registration race**: fold the investigate harness's `--delay` scenario into the suite — force a legacy extension's chrome registration ~30 s past startup and assert the chrome still comes live within a few seconds of the late registration. Empirical result: `autoRegister` registers chrome inline whenever it runs, so the loader never depends on the platform's startup re-scan window; this pins that behavior so a future loader refactor can't silently reintroduce the race.
- **Sweep unit test**: a `vm`-based, browser-free unit test for the `chrome.manifest*` startup sweep in BootstrapLoader.js — stub Services/IOUtils enough to run the sweep, assert stale uuid-named manifests and 0-byte leftovers are removed and a fresh manifest is written. CI-cheap in `pnpm test`, so the sweep isn't guarded only by the E2E leg.
[...]

PR #107, opened 2026-09-04
build: opt-in pre-push hook running the CI gates (pnpm hooks:install)

onemen:
## Summary
The repo ships one optional git hook: `githooks/pre-push` runs `pnpm lint && pnpm format && pnpm test` before a push leaves the machine — the same checks CI runs per PR — so a red CI run is predictable (the full hook costs ~11s cold, ~3–5s warm with caches).

- **Opt-in per ADR 0008** (which removed required hooks): `pnpm hooks:install` sets `core.hooksPath=githooks`; nothing changes for plain clones. It self-heals a **stale** `core.hooksPath` pointing at a non-existent directory (observed in the wild in this repo) and refuses to stomp a live one; equivalent spellings (`./githooks`, trailing slash, absolute) are normalized instead of erroring.
- Uninstall: `git config --unset core.hooksPath`. Single-push bypass: `git push --no-verify`.
- Gates are repo-wide (like CI), not scoped to the pushed range — scoping would let repo-wide breakage through. Docs-only contributors without a C toolchain push with `--no-verify` (documented; the `make analyze` leg hard-fails without gcc) and CI still runs the full gate.
- No generation steps in hooks: generated files stay on-demand per ADR 0008.
- Documented in `docs/DEVELOPING.md` → "Pre-push hook (opt-in)".

## Validation
- Hook exercised end-to-end: it ran all three gates and **caught an actually-unformatted file** in its own branch on the first run (then passed clean).
- eslint/prettier clean; 138 unit tests pass; gates re-verified on Node 24.20.0 (CI parity).
- `pnpm review:local`: 1 finding (hooksPath false positive on equivalent spellings) → fixed in this PR.

Complements #105 / #106 (hook gets much faster with those, but works independently).

Refs #30 (part of the pre-PR-gates tooling story).

🤖 Generated with Codebuff


onemen (commented):
Superseded — finding re-posted as a line-anchored thread on this PR.

tools/install-githooks.mjs
```diff
@@ -0,0 +1,86 @@
...
+        `Refusing to overwrite a live hooksPath. Inspect it, then unset manually:\n` +
+          `  git config --unset core.hooksPath`
+      );
+      process.exit(1);
+    }
+    console.log(`  replacing stale value with '${HOOKS_DIR_REL}'.`);
+  }
+}
+
+// Validate + chmod BEFORE flipping the config: a missing hook or a failed
```
onemen, line 64: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit `review:batch` run)

**Minor — right.** `core.hooksPath` was written before the hook was validated, so a missing `githooks/pre-push` or a failed `chmod` would have left git pointed at hooks that cannot run. Fixed in f415d27: validation and `chmod` now happen before the config write. Verified: with the hook removed, the installer exits 1 and the config is unchanged.

PR #116, opened 2026-09-04
fix(publish): link node_modules into --ref worktrees (Fixes #115)

### LINKED ISSUE #115: upload:local --ref crashes: temporary ref worktree has no node_modules
### Repro

```bash
pnpm upload:local -- --mode=prod --ref=<any commit>
```

Fails with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'minimatch' imported from
  …\Temp\firefox-scripts-ref-40a9ad4-…\tools\publish\gitignoreUtils.mjs
```

### Root cause

`runRefBuild()` (tools/publish/upload.mjs) creates a detached worktree at the ref and re-executes the ref's own `tools/publish/upload.mjs` inside it — deliberately, so an old ref is built by its own compatible publish scripts. But a fresh `git worktree add` carries no `node_modules` (gitignored, never materialized), so the first npm import in the ref's tooling crashes. Found today while building a `--ref=40a9ad4` snapshot; plain (non-`--ref`) `upload:local` is unaffected.

### Proposed fix

After `git worktree add`, link the current checkout's installed `node_modules` into the worktree (junction on Windows, symlink elsewhere) so the ref's scripts resolve dependencies from the existing install — keeping the re-exec-the-ref's-own-scripts design intact. Fall back to `pnpm install --ignore-scripts` in the worktree when there is no parent install to link. Remove the link before `git worktree remove` so the deletion can never traverse into the real store.

### Validation

`pnpm upload:local -- --mode=prod --ref=40a9ad4` (the exact failing command) completes and produces `dist/prod-main-40a9ad4/`… after the fix; parent checkout's `node_modules` verified intact afterwards.

_Discovered while investigating that the `latest` release predates the merged installer fixes (#106)._

onemen:
Fixes #115 — `pnpm upload:local -- --mode=prod --ref=<sha>` crashed on every run (`ERR_MODULE_NOT_FOUND: minimatch`) because `runRefBuild()` re-executes the ref's own `upload.mjs` inside a fresh detached worktree that has no `node_modules`.

## Change

- New `tools/publish/refNodeModules.mjs`: after `git worktree add`, link the current checkout's installed `node_modules` into the worktree (junction on Windows — no admin rights or developer mode needed; symlink elsewhere), and unlink it before worktree removal so the deletion can never traverse into the real store.
- `runRefBuild()` links right after creating the worktree; a parent checkout with no install now fails fast with a fix-it message instead of a cryptic module-not-found.
- The `--ref` design is unchanged: the ref is still built by its own compatible publish scripts; only dependency resolution is bridged.

## Validation

- The exact failing command now succeeds: `upload:local --mode=prod --ref=40a9ad4` → `dist/prod-40a9ad4-40a9ad4/` snapshot in 10.6s (this run built the pre-#106 ref that motivated the investigation).
- Parent checkout's `node_modules` verified intact afterwards; temp worktree removed.
- 5 new unit tests (`link / replace-stale / no-parent / unlink-safety / no-op`) — full suite 141 pass / 0 fail; `pnpm lint` ✓; `pnpm format` ✓.

## AI review (ADR 0020)

`pnpm review:local` outcome: <findings summary will be posted as review threads per the updated protocol once CI is green>.

Refs #115

### DIFF
--- a/tools/publish/refNodeModules.mjs
+++ b/tools/publish/refNodeModules.mjs
@@ -1,0 +1,50 @@
+#!/usr/bin/env node
+
+/**
+ * Dependency hand-off for `--ref` builds (tools/publish/upload.mjs →
+ * runRefBuild): a fresh `git worktree add` carries no node_modules (gitignored,
+ * never materialized), so the ref's own publish scripts — re-executed inside
+ * the worktree by design — crash on their first npm import.
+ *
+ * linkNodeModules() bridges that gap without touching the ref's sources: it
+ * exposes the current checkout's installed node_modules inside the worktree via
+ * a junction (Windows) or symlink (POSIX). Node resolves imports through the
+ * link, while the scripts themselves stay the ref's own compatible copies.
+ * Parent and worktree share one install; nothing is duplicated or installed
+ * anew. unlinkNodeModules() removes the link before the worktree is deleted, so
+ * the removal can never traverse into the real store.
+ */
+
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Point <worktree>/node_modules at <parent>/node_modules. Returns the link
+ * path, or null when the parent has no install to share (callers may fall back
+ * to running the package manager inside the worktree).
+ */
+export function linkNodeModules(parentRoot, worktreeRoot) {
+  const from = path.join(parentRoot, 'node_modules');
+  if (!fs.existsSync(from)) return null;
+  const to = path.join(worktreeRoot, 'node_modules');
+  fs.rmSync(to, {recursive: true, force: true});
+  if (process.platform === 'win32') {
+    // Junction: no admin rights, no developer mode, and removal never
+    // traverses into the target.
+    fs.symlinkSync(from, to, 'junction');
+  } else {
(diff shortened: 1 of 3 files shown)


tools/publish/refNodeModules.mjs
```diff
@@ -0,0 +1,52 @@
...
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Point <worktree>/node_modules at <parent>/node_modules. Returns the link
+ * path, or null when the parent has no install to share (callers may fall back
+ * to running the package manager inside the worktree).
+ */
+export function linkNodeModules(parentRoot, worktreeRoot) {
+  // Resolve up front: fs.symlinkSync interprets a relative target against the
```
onemen, line 27: 🤖 AI review triage (Codebuff agent — result of the local `review:local` run)

**Minor — right.** `fs.symlinkSync` resolves a relative target against the link's directory rather than the caller's cwd, so relative `parentRoot`/`worktreeRoot` inputs would have silently mislinked. Accepted: both paths are now `path.resolve`d up front, with a regression test. Fixed in this branch.

PR #117, opened 2026-09-04
fix(review): fail soft on non-object model replies (null crash)

onemen:
`tools/ai-review.mjs` crashed during a live `review:local` run (observed while reviewing #116):

```
TypeError: Cannot read properties of null (reading 'summary')
    at reviewFiles (tools/ai-review.mjs:446:56)
```

The per-file loop handles provider failure (`r.failed` → skip note) and invalid JSON (catch → skip note), but `JSON.parse("null")` **succeeds** — a provider that returns a bare `null` (or any non-object JSON) slipped past the catch and threw at `parsed.summary`, killing the entire run instead of skipping one file. The retry succeeded, which is why it never reproduced before.

## Change

After the parse, `parsed` that is not a non-null object is folded into the existing invalid-JSON path — same per-file "Model returned invalid JSON; findings skipped" summary, run continues. Missing `message.content` (`undefined`) lands on the same path via the parse throw.

## Tests

Four cases in `test/unit/publish/ai-review.test.mjs`, all asserting a skip summary + zero diagnostics instead of a throw: bare `null` (the observed shape), `42`, `"just a string"`, and missing `message.content`. Suite: 139 pass / 0 fail; lint ✓; format ✓.

🤖 This PR was created by an agent (Codebuff).

### DIFF
--- a/tools/ai-review.mjs
+++ b/tools/ai-review.mjs
@@ -438,7 +438,7 @@ export async function reviewFiles({
       parsed = JSON.parse(content);
     } catch {
       summaries.push(
-        `### \`${r.file}\` — ${r.providerName}\nModel returned invalid JSON; findings skipped.`
+        `### \`${r.file}\` — ${r.providerName}\nModel returned an invalid or unusable JSON reply; findings skipped.`
       );
       continue;
     }
@@ -448,7 +448,7 @@ export async function reviewFiles({
       // which then crashed the whole run at parsed.summary. Fail soft: same
       // per-file skip as invalid JSON.
       summaries.push(
-        `### \`${r.file}\` — ${r.providerName}\nModel returned invalid JSON; findings skipped.`
+        `### \`${r.file}\` — ${r.providerName}\nModel returned an invalid or unusable JSON reply; findings skipped.`
       );
       continue;
     }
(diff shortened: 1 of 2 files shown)


tools/ai-review.mjs
```diff
@@ -438,7 +438,17 @@ export async function reviewFiles({
       parsed = JSON.parse(content);
     } catch {
       summaries.push(
-        `### \`${r.file}\` — ${r.providerName}\nModel returned invalid JSON; findings skipped.`
+        `### \`${r.file}\` — ${r.providerName}\nModel returned an invalid or unusable JSON reply; findings skipped.`
+      );
+      continue;
+    }
+    if (typeof parsed !== 'object' || parsed === null) {
```
onemen, line 445: 🤖 AI review triage (Codebuff agent — result of the local `review:local` run)

**Minor — fair kernel, accepted.** For a bare `null` the JSON is valid, so the "invalid JSON" wording was technically misleading; both skip paths now read "invalid or unusable JSON reply" and the tests match. The companion finding on the test file was hedged speculation ("cannot confirm whether the function meets these expectations" without reading the implementation) — rejected as useless: the tests assert actual `reviewFiles` behavior and pass.

PR #134, opened 2026-09-05
feat(ci): url-watchdog meta-issue status dashboard + E2E validation column

onemen:
## What

The weekly URL watchdog (`tools/check-browser-downloads.mjs`) now maintains a single **`[url-watchdog] status` meta issue** as its dashboard, replacing the pile of per-release ledger issues:

- **Status table** for all 6 browsers: `Browser | Last verified | Size · SHA-256 | Last check | Status | Fallback (CI cache) | E2E validated` — short `[Sep 5](run-url)` links (never the full URL as text), status tags (✅ up to date / 🆕 new version / ❌ lookup failed / ⚠️ endpoint failed / 🔄 size changed / ⏳ first run), and a fallback column showing the last-known-good version CI's installer cache still serves when a check fails.
- **Version history** — the SHA-256 ledger, moved into the meta issue; grows only on runs with real version updates (capped at 10 entries).
- **Error issues only** now open (rot / size-change, exact-title dedup); new-version issues are no longer opened.
- **Auto-close**: a browser that checks green again closes its open failure issues with a "Resolved by watchdog run" comment.
- Meta issue is PATCHed only when content changed (no churn on no-op runs) and updated *before* the fail-closed exit, so failed runs still surface on GitHub while the baseline write stays fail-closed (ADR 0021).
- Baseline entries gain `checkedAt`/`checkedUrl` for the run links.

**E2E validation column** (task 3 of the request): the E2E `record-validation` job comments the validated firefox/firefox-dev versions on the meta issue (deduped — only when the versions change), and the watchdog workflow restores the `browser-validated` cache so the table shows `✅ 155.0.1` / `⏳ …` per hard-gate browser. e2e.yml gains `issues: write` for the comment.

## Validation

- `pnpm test` — 220 pass / 0 fail (10 new watchdog tests + 1 for the E2E column)
- `pnpm lint` — clean (eslint + C `gcc -fanalyzer`)
- `pnpm format` — clean
- Rendered meta-issue body eyeballed offline with current real data (shown in the conversation)

## Rollout

- Next weekly run creates the meta issue, seeded from the current baseline.
- The 5 now-superseded ledger issues (#124, #113, #114, #111, #100) get closed pointing at the meta issue.

## Summary by CodeRabbit

- **New Features**
  - Added a consolidated browser status view with release versions, check results, download details, checksums, and E2E validation status.
  - Added version history tracking for browser releases.
  - Added deduplicated error reporting that automatically closes issues after a successful check.
[...]

PR #137, opened 2026-09-05
fix(ci): record the updater legs' exact installed versions in E2E validation

onemen:
Fixes the deferred CodeRabbit finding from #134 (review finding 5): the E2E `record-validation` job re-resolved the *live* firefox/firefox-dev versions after a green run and recorded them as E2E-validated — but the updater matrix legs are path-filtered, so runs where those legs were skipped (core-only main pushes, or any `workflow_dispatch` — a path filter sees no commit diff on dispatch) recorded releases no leg ever tested, and the publish pre-flight (`--require-validated`) would then trust them.

## What changes

- **`test/e2e/shared/downloads.mjs`** — new `--installed-version` CLI mode: reads the installed binary's own version (`<binary> --version`, parsed by the exported pure `parseFirefoxVersion()`; falls back to `application.ini` next to the binary, which covers Windows GUI-subsystem builds that write no console output). Mozilla's "latest" redirect URLs embed no version, so the binary is the only ground truth of what a leg validated.
- **`.github/workflows/e2e.yml`** —
  - `updater` matrix legs capture that version after the E2E passes and upload a tiny per-leg artifact (`e2e-version-<browser>-<os>.json`). Matrix job outputs don't aggregate on GitHub (limitation #17245), hence files.
  - The `changes` job forces `updater=true` on a **full** manual dispatch (`browser` '' or 'all') — the single point that makes the snapshot/helper/updater/portable/browser-matrix jobs run, so the gate's applicability stays consistent. This is the post-release re-validation escape that was previously a no-op (a full dispatch ran no browser legs at all).
  - `record-validation` now `needs: [e2e-gate, updater]` and runs only when `needs.updater.result == 'success'` — i.e. only when the firefox/firefox-dev legs **actually ran and passed**. A core-only push (legs path-skipped) no longer writes a record.
- **`tools/ci/record-validated-versions.mjs`** — consumes the per-leg artifacts (new exported, unit-tested `collectLegVersions()`), requiring all three OS legs of a browser to agree on the version; no live `resolveVersion` anymore. Missing/disagreeing legs fail the job — a record with holes must not weaken the publish gate.
- **Docs** — `docs/ci-inventory.md` rows, `docs/DEVELOPING.md` (manual-escape/full-dispatch semantics, and the URL-watchdog bullet that still described the pre-#134 per-release ledger).

## Validation
- `pnpm test` 232 run / 229 pass / 0 fail (3 pre-existing skips)
- `pnpm lint` clean · `pnpm format` clean · `pnpm check:gates` holds
- 9 new unit tests: `parseFirefoxVersion` ×3 and `collectLegVersions` ×6 (+ existing file's count grew)

Note: an *existing* stale `browser-validated` cache entry could still back the publish gate until the next full green run; this PR makes the record trustworthy going forward.

## Summary by CodeRabbit

- **New Features**
  - Manual full-dispatch E2E runs now validate and record Firefox and Firefox Developer Edition versions.
  - Recorded versions are taken from the exact browser binaries installed during each operating-system test run.

- **Bug Fixes**
  - Version recording now requires all validated operating-system runs to complete successfully and agree on the installed versions.
  - Improved handling and validation of installed browser version detection.

- **Documentation**
  - Updated CI and development guidance for manual validation runs and browser version tracking.

- **Tests**
  - Added coverage for browser version parsing and validation artifact handling.


onemen:
@coderabbitai review

PR #138, opened 2026-09-05
fix(ci): correct the meta-issue comment API path in record-validation

onemen:
Fixes a bug surfaced by the first live full-dispatch record after #137: `notifyMetaIssue` (in `tools/ci/record-validated-versions.mjs`) built its comment URLs as `/repos/{issue-number}/comments` — the issue number in the *repository* slot — instead of `/repos/{owner}/{repo}/issues/{n}/comments`. Every attempt 404'd:

```
meta issue notification failed (non-fatal): GitHub API GET /repos/136/comments?...: HTTP 404 Not Found
```

So the "Validated by E2E: firefox=X · firefox-dev=Y — [run](…)" comment never landed on the `[url-watchdog] status` meta issue (the notification is non-fatal, which is why the job and record still succeeded).

Fix: both the GET (dedup check) and POST now use `/repos/${repo}/issues/${meta.number}/comments`, matching the watchdog's own GitHub calls in `check-browser-downloads.mjs`.

Verified: lint, format, and the record-validated-versions unit tests pass.

### DIFF
--- a/tools/ci/record-validated-versions.mjs
+++ b/tools/ci/record-validated-versions.mjs
@@ -208,16 +208,19 @@ async function notifyMetaIssue(record) {
       console.log('meta issue not found yet — validation comment deferred to the watchdog');
       return;
     }
+    // The comment endpoints are scoped under /repos/{owner}/{repo}/issues/{n}
+    // (issue comments, not repository comments) — matching the watchdog's own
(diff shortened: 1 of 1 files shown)


onemen:
@coderabbitai review

PR #141, opened 2026-09-05
Installer test surface: --port 0 / --server-only / --env-file + E2E layer

### LINKED ISSUE #129: Installer test surface: --port 0 / --server-only flags + env.json manifest
Part of #3 (Phase 4 — E2E matrix). Source: test-infrastructure audit absorbed into `docs/future-work.md` §6 (PR #128).

The installer binary is currently only testable on `DEFAULT_PORT=8777` with a full browser scan, which blocks parallel CI jobs and headless coverage of the second-instance path.

Scope (`installer/src/main.c` + `installer/Makefile`):

- `--port 0` — bind an OS-ephemeral port so parallel CI jobs/racks never collide on 8777 (free-port + no-hijack contract assertions).
- `--server-only` — skip the browser scan so the second-instance detection path (`tcp_listening`) is reachable headless.
- `env.json` deployment manifest — the running installer writes port/URLs/hashes/run-id to a file the E2E harness reads (local/CI parity, no port scraping).

Unblocks: installer API-contract tests, free-port/no-hijack/second-instance coverage, and any future Phase 4 leg that needs a headless installer.

onemen:
Closes #129. Part of #3 (Phase 4 — E2E matrix).

## What

The installer binary was only testable on `DEFAULT_PORT=8777` with a full browser scan, blocking parallel CI jobs and headless coverage of the second-instance path. This PR adds the three test-surface flags from the issue and an E2E layer that exercises them.

### Installer (`installer/src/main.c`)

- **`--port <n>`** — bind an explicit port; `--port 0` binds an OS-ephemeral port. `http_server_start` already reported the actual bound port via `getsockname`, so the UI URL is rebuilt from it. When `--port` is given, the default-port hijack probe (`tcp_listening(DEFAULT_PORT)` → open the other instance's tab) is skipped: parallel test runs must never be redirected to somebody else's UI.
- **`--server-only`** — skip the browser scan, never open a UI tab, and (added after the first E2E run caught it) never open a tab via the second-instance handoff paths. Implies `--smoke-test` rules so it never aborts on zero browsers.
- **`--env-file <path>`** — after the server binds and the token exists, write a JSON manifest `{port, token, runId, uiUrl}`. The harness reads this instead of scraping stdout or assuming the port. Best-effort: a manifest failure never breaks a working server.

The runtime server flags (`--verbose`, `--smoke-test`, `--server-only`, `--port`, `--env-file`) now parse over **every** argument so they compose (`installer --server-only --port 0 --env-file ...`). The one-shot modes (`--help`, `--test-hash`, `--test-self-update`, `--scan-only`) still dispatch on `argv[1]` as before.

### E2E (`test/e2e/installer/installer-e2e.mjs`)

New **test-surface layer**, on by default (`--no-test-surface` skips):

- **TS-1**: `--port 0` + `--env-file` → manifest written, port is ephemeral and ≠ 8777, token is 16-hex, `uiUrl` consistent, `/api/ping` answers on the ephemeral port.
- **TS-2**: `--port <fixed>` binds exactly that port; manifest reports it; server answers.
- **TS-3**: a plain second installer (no `--port`) while one serves the default port reports the running installer, exits non-zero, writes no manifest, and the first server keeps answering. No browser tab is opened (the `--server-only` hermeticity guard).
- **TS-4**: `--server-only` output contains no browser-scan line and announces the skipped scan.

**Windows SO_REUSEADDR note** (learned the hard way, encoded in the test): a second bind on an *explicitly* busy port can **succeed** on Windows, so "second instance refuses explicit busy port" is not a portable contract. The portable second-instance behavior is the default-port `tcp_listening` probe — TS-3 exercises exactly that, using `--server-only` without `--port` (the production shape, headless).

### Docs

`docs/DEVELOPING.md` → Installer E2E section documents the new layer and flags.

## Validation

- `make dist_win MODE=dev` builds clean; `make analyze` (gcc -fanalyzer) no findings
- E2E locally on Windows: **50/50 passed** (30 HTTP layer + 14 test-surface + UI-independent); `--no-test-surface` still 35/35
- `pnpm lint` / `pnpm format` / `pnpm test` (229 pass) green

CI will exercise the same layer on ubuntu/macos/windows via the existing `installer` job (no workflow change needed — the layer runs by default).

Fixes #129


installer/src/main.c
```diff
@@ -56,6 +66,44 @@ static char g_ui_host_profile[MAX_PATH_LEN] = "";
...
+// harness needs; it is NOT a monotonic clock reading) — lets the E2E harness
+// tell two installer runs apart when both write env.json manifests.
+static unsigned long long run_id_ms(void) {
+#ifdef _WIN32
+    FILETIME ft;
+    GetSystemTimeAsFileTime(&ft);
+    unsigned long long t = ((unsigned long long)ft.dwHighDateTime << 32) | ft.dwLowDateTime;
+    return (t - 116444736000000000ULL) / 10000ULL;  // 100ns ticks -> ms since epoch
+#else
+    struct timespec ts;
```
onemen, line 84: 🤖 AI review triage (Codebuff agent — result of pnpm review:local on this PR)

Accepted: the comment claimed 'monotonic run id' while run_id_ms() reads the wall clock. Comment fixed in 477e4ff — the harness only needs cross-run uniqueness.

test/e2e/installer/installer-e2e.mjs
```diff
@@ -284,6 +292,234 @@ async function runHttpLayer(counter, sessionToken) {
...
+    proc.stderr.on('data', d => (out += d.toString()));
+    const started = Date.now();
+    const poll = setInterval(() => {
+      let manifest = null;
+      try {
+        manifest = JSON.parse(fs.readFileSync(envFile, 'utf8'));
+      } catch {
+        // not written yet (or process refused to start)
+      }
+      const exited = proc.exitCode !== null || proc.signalCode !== null || proc.killed;
```
onemen, line 316: 🤖 AI review triage (Codebuff agent — result of pnpm review:local on this PR)

Accepted: spawnWithEnvFile only treated exitCode/killed as exited, so a signal kill polled the full timeout. proc.signalCode is now checked too (477e4ff).

onemen: Fixed in 477e4ff — resolving.

installer/src/main.c
```diff
@@ -2485,19 +2580,27 @@ static int main_impl(int argc, char *argv[]) {
...
         return 1;
     }
 
     // Start HTTP server.  The UI tab ALWAYS opens: it is the only component
     // with network access, so even when everything is up-to-date it must run
     // to fetch + ingest the manifest and confirm the status to the user.
-    int port = http_server_start(DEFAULT_PORT);
+    // --port <n> overrides the compiled default; 0 binds an OS-ephemeral port
+    // (getsockname reports the actual port, so the UI URL is rebuilt from it).
+    int port = http_server_start((unsigned short)(g_requested_port >= 0 ? g_requested_port
```
onemen, line 2602: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**minor — bind-failure path reports the default port, not the requested one**

With `--port <n>` (`n` ≠ 8777), if binding `n` fails, the failure branch below this call still prints `Could not start HTTP server (port %d in use)` with `DEFAULT_PORT` and opens `http://localhost:8777/` as the "another installer is running" fallback — naming a port the server never tried and pointing the tab at a possibly unrelated instance. Build both the error message and the fallback URL from the effective requested port (`g_requested_port >= 0 ? g_requested_port : DEFAULT_PORT`), consistent with the `http_server_start` call on this line.

onemen: Fixed in bd55ee7 — resolving.

PR #142, opened 2026-09-05
E2E: snap Firefox Linux leg (GreD verification + installer/updater)

### LINKED ISSUE #55: E2E on Linux with and without snap (snap Firefox GreD path is unverified)
Part of issue #3, advisory (availability depends on the runner having snapd + root to install the snap).

**Gap:** CI Linux legs install the official Mozilla tarball directly (download.mozilla.org). The Ubuntu-default snap packaging is never exercised, and `docs/future-work.md` §2.1 has an **open** item: snap Firefox's GreD — documented as `/etc/firefox` — is unverified ("GreD differs from the documented target — verify actual path"). Detection is process-scan + lock-file + `application.ini`-beside-the-binary, which should work under confinement, but nothing proves it.

**Proposed coverage:**
- [x] Verify the actual GreD path of snap-installed Firefox on an Ubuntu runner (install the snap with elevated privileges, or the `--classic` variant if needed) and reconcile `docs/future-work.md` §2.1.
- [x] Installer E2E (HTTP/UI layer) against snap Firefox: detection, version, GreD discovery.
- [x] Updater E2E against snap Firefox (Linux leg): tab opens, config install into the snap GreD (elevation is manual-only on hosted runners — assert what *can* run).
- [x] Non-snap (deb/rpm-style) comparison leg if cheap — the current tarball leg already covers a portable-style layout. Not added as a separate leg: Ubuntu's apt `firefox` is itself a snap transitional package, so the existing tarball legs remain the non-snap comparison.

**Where:** new leg(s) in `.github/workflows/e2e.yml` + recipes in `test/e2e/shared/downloads.mjs`; note that hosted runners need root for snap — gate on availability, don't hard-fail the matrix.

**Delivered by PR #142** (merged `b83d2e3`): `snap-firefox` job on ubuntu-latest installs the snap build, reports `findGreDir()` vs the actual snap layout in the step summary and hard-fails on divergence, and runs the installer + updater E2E against snap Firefox. Gated advisory on a snapd probe per the "Where" note; §2.1 reconciled in the same PR.

onemen:
Closes #55. Part of #3 (Phase 4 — E2E matrix). Stacked on #141 (shares the installer E2E test-surface layer; rebase to `main` after #141 merges).

## What

The CI Linux legs always installed the official Mozilla tarball (download.mozilla.org), so the Ubuntu-default **snap** packaging was never exercised and snap Firefox's GreD stayed an **open** item in docs/future-work.md §2.1 ("GreD differs from the documented target — verify actual path"). This PR adds the missing leg.

### test/e2e/shared/downloads.mjs — firefox-snap recipe (Linux only)

- Installs via the snap store (`snap install firefox --classic`; the CI step runs it with sudo — hosted runners have root).
- Returns /snap/bin/firefox explicitly (the stable launcher path).
- --url reports a stable snap://firefox origin — the store resolves the revision at install time, so there is no cacheable tarball URL (no cache step in the job).
- Firefox stable only: Dev Edition has no snap channel.

### .github/workflows/e2e.yml — new snap-firefox job (ubuntu)

Per the issue's "gate on availability, don't hard-fail the matrix":

1. **Runtime snapd probe** — the whole leg skips quietly (with a ::warning::) when snapd is unavailable.
2. **GreD verification — the point of the leg** — runs findGreDir() against /snap/bin/firefox, checks /etc/firefox existence, writes the probe table (binary, derived GreD, mount) to the step summary, and **hard-fails if findGreDir() returns a non-existent directory** — so browsers.mjs can never silently drift from the actual snap layout.
3. **Installer E2E** (HTTP + UI + the #129 test-surface layer) against the snap binary — detection must find and report /snap/bin/firefox.
4. **Updater E2E** against the snap binary — tab opens, config seeds, status renders. Elevation into the snap GreD stays on the manual release checklist (§2), as the issue specifies ("assert what *can* run").

The job is **advisory** in the E2E gate (like the fork browser-matrix).

### .github/actions/verify-gate/verify.sh

An advisory job may now **skip itself for availability** without a gate warning (snap-firefox's snapd probe): success and skipped both count as OK for advisory legs. A deliberate availability skip is not a degradation.

### Docs

- docs/future-work.md §2.1: the snap row's "**open**: verify actual path" is closed — it now describes the verification mechanism (CI leg + step-summary probe + hard-fail on divergence); elevation stays manual.
- docs/ci-inventory.md: new job row (advisory, runtime availability gate).
- docs/DEVELOPING.md: CI matrix row.
- e2e.yml applicability block: comment warning that the gate-coverage checker parses bare 12-space name: pairs only (an inline comment inside the block silently dropped the following entry — caught by pnpm check:gates).

## Validation

- pnpm check:gates ✓ (11 jobs, gate covers all — including the new one)
- pnpm lint ✓ (incl. gcc -fanalyzer), pnpm format ✓, pnpm test ✓ (229 pass)
- resolveDownloadUrl('firefox-snap', 'linux') → snap://firefox; win/mac throw the standard manual-only error; firefox/firefox-dev tarball recipes untouched
- The snap install + GreD probe + E2E legs themselves can only run on a Linux runner with snapd — CI validates them; this Windows machine cannot

Fixes #55


installer/src/main.c
```diff
@@ -1169,6 +1169,9 @@ int handle_api_install(int client_fd, const char *query, const char *body, size_
 
     strncpy(g_binary_dir, detected_browsers[browser_idx].binary_path, MAX_PATH_LEN);
     get_parent_dir(g_binary_dir);
+    // Snap installs keep their autoconfig in /etc/firefox, not the read-only
+    // /snap/... app dir — fx-folder must be copied there.
+    config_dir_for_app_dir(g_binary_dir, g_binary_dir, sizeof(g_binary_dir));
```
onemen, line 1174: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**major — overlapping src/dst in `snprintf` (undefined behavior in C)**

`config_dir_for_app_dir(g_binary_dir, g_binary_dir, sizeof(g_binary_dir))` passes the same buffer as source and destination. In the helper's non-snap branch (`installer/src/detect_browser.c:1054`) that runs `snprintf(out, out_sz, "%s", app_dir)` with `out == app_dir` — copying between overlapping objects is UB per the C standard. glibc happens to tolerate the self-copy today, but the pattern is a footgun (UCRT/sanitizers can flag it) and this call site is the only one that aliases the buffers.

Suggested fix — helper-side, minimal: the snap branch writes a literal (no overlap), so only the fall-through copy needs guarding:

```c
if (app_dir && strstr(app_dir, "/snap/")) {
    snprintf(out, out_sz, "/etc/firefox");
    return;
}
if (out == app_dir) return; // self-copy no-op — avoids overlapping snprintf
snprintf(out, out_sz, "%s", app_dir ? app_dir : "");
```

.github/workflows/e2e.yml
```diff
@@ -487,6 +487,168 @@ jobs:
...
+          echo "| --- | --- |" >> "$GITHUB_STEP_SUMMARY"
+          echo "| FIREFOX_BINARY | \`$BIN\` |" >> "$GITHUB_STEP_SUMMARY"
+          echo "| findGreDir() | \`$GRED\` |" >> "$GITHUB_STEP_SUMMARY"
+          if [ -d /etc/firefox ]; then
+            echo "/etc/firefox exists → documented GreD target confirmed for this snap build"
+            echo "| /etc/firefox exists | yes |" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "/etc/firefox does NOT exist — findGreDir() must be reconciled with the actual path"
+            echo "| /etc/firefox exists | no |" >> "$GITHUB_STEP_SUMMARY"
+            echo "| snap mount | \`$(snap run --shell firefox -c 'echo $SNAP' 2>/dev/null || n/a)\` |" >> "$GITHUB_STEP_SUMMARY"
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**minor — `|| n/a` runs a nonexistent command**

When the `snap run --shell firefox -c 'echo $SNAP'` probe fails, the fallback `|| n/a` executes a command literally named `n/a`: the shell reports "command not found" into the step log and the `$GITHUB_STEP_SUMMARY` table cell stays empty instead of showing the intended fallback text. Use `|| echo n/a` so the cell reads `n/a` and no spurious error is emitted.

installer/src/main.c
onemen: Fixed in 5476137 — resolving.

.github/workflows/e2e.yml
onemen: Fixed in 5476137 — resolving.

Issue #143, opened 2026-09-06
Reliable browser downloads: progress-aware downloadTo + watchdog→E2E auto-dispatch


onemen:
## Status: COMPLETE (closed 2026-09-06)

| Item | State |
| --- | --- |
| PR 1 — progress-aware `downloadTo` | ✅ [#144](https://github.com/onemen/firefox-scripts/pull/144) merged (`f6ecb4c`) |
| PR 2 — watchdog auto-dispatch | ✅ [#145](https://github.com/onemen/firefox-scripts/pull/145) merged (`f130df8`) |
| Follow-up — download-time column + knobs docs | ✅ [#146](https://github.com/onemen/firefox-scripts/pull/146) merged (`4d61f4f`) |
| Workflow fix — escape dispatch needs snapshot (found by the proof) | 🔶 [#147](https://github.com/onemen/firefox-scripts/pull/147) open |
| Live verification | ✅ [simulated-release proof, final comment](https://github.com/onemen/firefox-scripts/issues/143#issuecomment-5559366177) |

The next *real* browser release is a passive confirmation of the full loop, not an open work item.

---

## Summary

Two gaps in the browser-download chain, both confirmed with live measurements on 2026-09-06:

1. **`downloadTo` (test/e2e/shared/downloads.mjs) kills slow-but-healthy downloads.** Each attempt is bounded by a wall-clock `AbortSignal.timeout(300_000)` covering headers + the entire body. There is no streaming (whole body is buffered via `arrayBuffer()`), no progress awareness, and no resume — a retry restarts from byte 0.
2. **The URL watchdog records new browser versions but dispatches no E2E.** `docs/ci-inventory.md` says the watchdog "should dispatch targeted updater compatibility E2E for that browser", and the publish drift-gate error text tells operators to rely on it — but the trigger was never implemented.

## Evidence (2026-09-06)

**Home link (the operator's morning repro, dl.librewolf.net → 155.0.1-1 portable zip):**

- 165,930,429 B (158.2 MB) downloaded in **335 s** — avg **484 KB/s**, 10 s windows ranging 311–723 KB/s, TTFB 410 ms.
- SHA-256 matched the vendor `.sha256sum` byte-for-byte → **healthy-but-slow, not corruption**.
- Repro of the repo's exact per-attempt mechanism (`fetch(url, {signal: AbortSignal.timeout(300_000)})`, what `fetchWithRetry` does):

  ```
  headers ok after 0.3s, streaming body...
[...]

PR #144, opened 2026-09-06
fix(e2e): progress-aware browser downloads — stall detection + Range resume

onemen:
Part of #143.

## Problem

`downloadTo` bounded each attempt with a flat wall-clock `AbortSignal.timeout(300_000)` covering headers **and the whole body**, buffered the entire response via `arrayBuffer()`, and restarted from byte 0 on every retry. Measured on 2026-09-06 (issue #143): the same 158 MB LibreWolf installer takes **13 s from CI runners** but **335 s (5 min 35 s) over a home link** at ~484 KB/s — and the repo's exact per-attempt mechanism aborted that healthy transfer at exactly 300 s with ~130 MB transferred, then would have retried from zero five times (~25 min) before failing.

## Change

- `downloadTo` **streams to disk** and aborts only when **no bytes advance for the stall window** (default 60 s) — a 0.3 MB/s trickle delivers a chunk every ~2 s and is never killed; only a genuinely dead stream trips it.
- A generous **total budget** (default 20 min — 3.5× the worst observed home case, 90× the worst CI case) bounds the overall download; the per-attempt `AbortSignal` now enforces the *remaining budget share*, not a flat 5 min.
- Interrupted attempts **resume via `Range: bytes=<written>-`** and append; a server without range support answers `200` and the attempt restarts cleanly; the final size is validated against `Content-Length` when known. The partial file survives across attempts (flushed on error).
- Stall window, budget, and retry backoff are **env-overridable** (`DOWNLOAD_STALL_TIMEOUT_MS` / `DOWNLOAD_TOTAL_BUDGET_MS` / `DOWNLOAD_RETRY_BACKOFF_MS`), so the unit tests exercise the behavior in milliseconds — same pattern as the resolver's `BROWSER_RESOLVER_BACKOFF_MS`.
- Progress heartbeat every 10 MB in the CI log (rate in KB/s), so a slow transfer is visibly alive instead of looking hung.
- Watchdog `check` job: `timeout-minutes: 30` so the transfer budget fits the job (corrects the `10` suggested in the ci-inventory follow-up row).
- HEAD cache-reuse path unchanged (CI cache restore still avoids the download entirely).

## Tests

Three permanent unit tests (local `node:http` servers, sub-second) pin the new contract:

1. **Stalled connection fails fast** — first chunk then silence: rejects after the stall window (not a wall clock), partial file kept for a later resume.
2. **Slow-but-progressing completes** — bytes dripping every 60 ms: byte-complete file (the #143 home-link case).
3. **Interrupted transfer resumes via Range** — attempt 1 cut mid-body, retry sends `Range: bytes=1024-`, server answers `206`, final file byte-complete.

`pnpm test` 232 pass / 0 fail (3 skipped, pre-existing platform skips); `pnpm lint` clean (incl. gcc -fanalyzer); `pnpm format` clean; pre-push gate re-ran the full suite.

## Docs

`docs/ci-inventory.md`: watchdog timeout row → `timeout-minutes: 30` (implemented); new-version row notes the progress-aware transfer.


.github/workflows/url-watchdog.yml
```diff
@@ -44,6 +44,12 @@ jobs:
     name: check browser download URLs
     if: github.event_name != 'pull_request' && github.event_name != 'merge_group'
     runs-on: ubuntu-latest
+    # The full-download path is progress-aware (issue #143): a slow vendor
+    # download runs to its 20 min transfer budget (plus retries) instead of
+    # being killed early. The job timeout must fit that budget, not the other
+    # way around.
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
```
onemen: 🤖 **AI review triage (Codebuff agent — result of the CodeRabbit review:batch run over #144+#145)** — critical — accepted. Duplicate `runs-on: ubuntu-latest` on the job (line 46 already declares it): a duplicate mapping key makes GitHub reject the whole workflow file, taking the URL watchdog down entirely. Fix: drop the duplicate.

test/e2e/shared/downloads.mjs
```diff
@@ -221,36 +222,138 @@ async function installTarball(url, browser, dest = path.join(os.homedir(), 'fire
...
+      try {
+        const result = await Promise.race([reader.read(), stall.promise]);
+        if (result.done) break;
+        chunk = result.value;
+      } finally {
+        stall.cancel();
+      }
+      written += chunk.length;
+      if (!stream.write(Buffer.from(chunk))) {
+        await new Promise(resolve => stream.once('drain', resolve));
```
onemen: 🤖 **AI review triage (Codebuff agent — result of the CodeRabbit review:batch run over #144+#145)** — major — accepted. Awaiting `drain` alone can hang forever if the WriteStream errors while backpressured: the error listener set at stream creation only flips a flag that is read on the *next* chunk, which never arrives. Fix: race `drain` against `error` so the promise always settles.

.github/workflows/url-watchdog.yml
onemen: 🤖 Fixed in 90f8481 (duplicate runs-on removed; gates re-ran green locally) — resolving.

test/e2e/shared/downloads.mjs
onemen: 🤖 Fixed in 90f8481 (drain raced against error; all three transfer tests still green). No deterministic unit test added — forcing an ENOSPC-style mid-backpressure stream failure is not feasible cross-platform; the surrounding behavior stays pinned by the stall/drip/resume tests. Resolving.

PR #145, opened 2026-09-06
feat(watchdog): dispatch browser E2E when a new release is recorded

onemen:
Part of #143.

## Problem

The URL watchdog recorded new browser versions, updated the `[url-watchdog] status` meta issue — and dispatched nothing. `docs/ci-inventory.md` says the watchdog "should dispatch targeted updater compatibility E2E for that browser", and the publish drift-gate error text tells operators to rely on it, but the trigger was never implemented. Demonstrated live by [run 34022364893](https://github.com/onemen/firefox-scripts/actions/runs/34022364893) on 2026-09-06: it recorded librewolf `155.0-1 → 155.0.1-1` and zen `1.21.16b → 1.22b` and dispatched no E2E — the new releases sit untested until an unrelated `core`/`updater` push or a manual dispatch.

## Change

- After the baseline persists (so a fail-closed run dispatches nothing), `new-version` findings map to E2E dispatches through the pure, unit-tested **`planDispatches(findings)`**:
  - **Forks** (librewolf/floorp/zen/waterfox): one single-browser dispatch per browser — the `browser` input collapses the matrix to that leg, the ADR 0021 manual escape, whose runs sit in their own non-cancelled concurrency group.
  - **Hard gates** (firefox/firefox-dev): **one** full dispatch (`browser=all`) covering both — it runs the updater legs on all 3 OSes plus `record-validation`, refreshing the validated-versions record the publish gate reads. Never two: a second full dispatch would land in the same cancel-in-progress concurrency group and kill the first before record-validation runs.
  - `first-run` findings dispatch nothing (a cache eviction is not a release); repeated forks dedupe; the fork set caps the escapes.
- A failed dispatch is a `::warning::` annotation, not a failed run — the baseline and issue surface already succeeded, and the next run re-triggers only if the version is still new.
- Workflow: job-level permissions on `check` (`contents: read`, `issues: write`, `actions: write` — the E2E workflow's scoping precedent) and an explicit minimal block on the stateless `pr-check`; the workflow-level block is unchanged.

## Tests

Six new unit tests pin the planner: per-fork escapes, hard gates → exactly one full dispatch (with the concurrency rationale), mixed release, `first-run` → `[]`, rot/size-change → `[]`, dedupe + fork-set cap. `pnpm test` 235 pass / 0 fail; `pnpm lint`, `pnpm format` clean.

## Docs

`docs/ci-inventory.md`: the "should dispatch" paragraph now describes the implemented behavior (forks single-browser, hard gates one full dispatch, post-baseline only, failure = warning) and the new-version row adds the dispatch cell.

Depends on nothing in #144, but lands naturally after it: dispatched legs then download the new release with the progress-aware downloader.


tools/check-browser-downloads.mjs
```diff
@@ -1056,6 +1108,32 @@ export async function main() {
...
   }
 
+  // Auto-dispatch the browser E2E for new releases (issue #143): docs/ci-
+  // inventory.md promised this ("should dispatch targeted updater compatibi-
+  // lity E2E") but it was never wired up — new releases sat untested until an
+  // unrelated push or a manual dispatch. Runs only after the baseline persis-
+  // ted (the fail-closed exit above already returned otherwise), so a browser
+  // is dispatched at most once per recorded version. A failed dispatch is a
+  // warning, not a failed run — the baseline and the issue surface already
+  // succeeded, and the next run re-triggers only if the version is still new.
```
onemen: 🤖 **AI review triage (Codebuff agent — result of the CodeRabbit review:batch run over #144+#145)** — major — accepted. Correct: the baseline is already written before the dispatch block, so a failed dispatch is never retried (the next run sees `ok` and produces no new-version finding); the comment — and the PR description — claimed the opposite. Fix: honest comment, plus a deduped failure issue (auto-closed once the browser checks green again) so a failed dispatch does not leave the release silently untested.

onemen: 🤖 Fixed in 008fb15 (honest comment + deduped per-browser failure issue with the manual retry command; issue title deliberately outside the auto-close set since the watchdog cannot verify an E2E run happened). Resolving.

PR #147, opened 2026-09-06
e2e: single-browser escape dispatch must build the snapshot

onemen:
## Why

The #143 close-out proof exposed a real workflow bug. The watchdog auto-dispatch for a simulated floorp release ([run 34033626251](https://github.com/onemen/firefox-scripts/actions/runs/34033626251), `browser=floorp`) completed in 40 s with **every E2E leg skipped**:

- `browser-matrix`'s own `if:` correctly treats a single-browser dispatch as applicable;
- but `browser-matrix` **needs `snapshot`**, and `snapshot`'s `if:` only knew `updater/core` path-filter outputs — on a dispatch the path filter sees no commit diff, so snapshot skipped;
- a needs-chain skip is **transitive** → browser-matrix skipped despite its condition.

The same latent bug sat under `pnpm ci:download`'s manual escape (ADR 0021) since it was written — no partial dispatch had ever run before, so it was invisible.

## Fix

- `snapshot` treats a single-browser `workflow_dispatch` as applicable (same combined condition style as `browser-matrix`).
- The `e2e-gate` `applicability:` mirror updated identically.
- `tools/check-gate-coverage.mjs` pins the new combined `if:` for both jobs (gate contract stays enforced).
- `docs/ci-inventory.md`: snapshot row updated.

## Proof

Dispatched `e2e.yml` with `browser=floorp` **from this branch** ([run 34034280784](https://github.com/onemen/firefox-scripts/actions/runs/34034280784)): snapshot ran, the browser-matrix collapsed to exactly one `updater E2E · floorp · windows-latest` leg, and the whole run concluded **success**. (The other required legs also ran on this branch-only dispatch because the workflow diff itself made the path filter match — incidental, and they passed.)

Part of #143

🤖 Generated with Codebuff


docs/ci-inventory.md
```diff
@@ -13,7 +13,7 @@ change.
...
-| E2E / build dev snapshot                         | Builds shared updater snapshot                                                                                                                                     | Same E2E triggers                                      | Runs when `updater` or `core` is true                                                                                                                                                                         |
+| E2E / build dev snapshot                         | Builds shared updater snapshot                                                                                                                                     | Same E2E triggers                                      | Runs when `updater` or `core` is true; a single-browser dispatch also forces it (browser-matrix needs the snapshot — a needs-chain skip is transitive)                                                        |
```
onemen, line 16: 🤖 AI review (Codebuff agent — pnpm review:local, Gemini) — warning — two table rows were merged onto one line by the snapshot-row edit: one 8-cell row that GitHub renders by dropping the snapshot row from the table. Verified right (row count on this branch: 1 instead of 2, vs origin/main). Fixed in 8bd4383.

onemen (commented):
🤖 AI review (Codebuff agent — pnpm review:local, Gemini; CodeRabbit batch still rate-limited, will use the slot after reset only if still wanted)

1 finding, triaged RIGHT and fixed in 8bd4383:

- docs/ci-inventory.md (table rows) — the snapshot-row edit had merged two table rows onto one line (a malformed 8-cell row; GitHub would render it by dropping the snapshot row from the table). The formatting gate passed because prettier treats it as a single valid row. Split back into two rows; prettier re-aligned.

Note: e2e.yml and check-gate-coverage.mjs were reviewed clean by the local pass; the workflow's needs-chain semantics are additionally proven live by dispatch run 34034280784 (snapshot ran, matrix collapsed to the floorp leg, run success).

PR #151, opened 2026-09-06
Guard against AV false positives on installer_win.exe (PE metadata + scan gate + stable toolchain)

onemen:
## Problem
The Sep 5 release artifact `installer_win.exe` was flagged by Windows Defender as
`Program:Script/Wacapew.A!ml` (a **machine-learning** heuristic). Investigation proved it is a
false positive specific to that artifact's exact bytes:
- Every local rebuild of the same source (previous-release, exact Sep-5-era, and current) scans **clean** with today's Defender.
- The CI-built **dev** binary from the same publish window (same toolchain, ~512 bytes different config) is also clean.
- The confirmed trigger is a **toolchain bump, not a code change**: the publish workflow ran `msys2/setup-msys2` with `update: true`, and a full `pacman -Syu` on 2026-09-05 pulled the gcc **16.1.0 → 16.2.0** package update (published to MSYS2 repos the day before). The recompiled bytes landed inside the `!ml` model's detection pocket.
- Multiple engines show the classic unsigned-binary ML-FP profile (McAfee local/cloud discrepancy, Bkav consistent; Jotti/Virscan clean). `helper_win.exe` was never flagged — it already carries a version resource; `installer_win.exe` had none.

## Fixes
1. **PE metadata** — add `installer/src/installer.rc` + `installer.manifest` (version resource, asInvoker, DPI-aware, Win10/11 compat), compiled by `windres` and linked on the Windows build only (`WIN_RES` empty on Linux/mac). A stripped unsigned PE with no version info is the #1 ML-FP profile.
2. **AV scan gate at publish time** — new `tools/scan-av.mjs`, wired into `upload.mjs`: scans the **exact staging bytes about to be uploaded** (Windows: Defender `MpCmdRun`; Linux/macOS: ClamAV) and **refuses to publish on any detection**. A missing engine only warns (GitHub Windows runners often run Defender in passive mode). Plus `pnpm scan:av` for local checks.
3. **Stable toolchain** — drop `update: true` on the publish workflow so gcc changes are deliberate, not pulled mid-publish.
4. **Docs** — root cause + the scan/verify workflow + Microsoft WDSI false-positive submission link in `docs/DEVELOPING.md`.

## Validation
- `pnpm lint` (eslint + clang-format + `gcc -fanalyzer` + markdownlint), `pnpm format`, `pnpm test` — all green.
- `pnpm scan:av -- dist/installer/installer_win.exe` → clean; EICAR test file → correctly detected (exit 1).
- The Windows exe now shows proper version metadata (FileDescription/ProductName/CompanyName 1.0.0).

## Follow-ups (not in this PR)
- Re-publish `latest` once merged — the gate scans the exact CI artifact before upload.
- Optional: submit the flagged hash `76109ec3…` to Microsoft WDSI as a false positive to clear it for existing users.
- Optional: SignPath Foundation free OSS code signing as the durable layer.

### DIFF
--- a/docs/DEVELOPING.md
+++ b/docs/DEVELOPING.md
@@ -167,10 +167,10 @@ The publish flow also runs every built binary through VirusTotal when `VT_API_KE
 (GitHub secret on CI; root `.env` for a local `pnpm upload`). The publish fails when ≥
 `VT_FAIL_THRESHOLD` (default 3) engines report a binary as malicious **or** when a veto engine
 (`VT_VETO_ENGINES`, default `Microsoft`) reports it as malicious at any count — a Microsoft/Defender
-verdict must never ship, even alone. 1–2 hits from non-veto engines warn but do not block (the
-known-FP band). The run log names the flagging engines. Without a key it just skips with a warning,
-and an analysis VirusTotal has not finished when the poll times out is reported as a skip — never as
-clean.
+verdict must never ship, even alone. Hits below the configured threshold from non-veto engines warn
+but do not block (the known-FP band at the default of 3). The run log names the flagging engines.
+Without a key it just skips with a warning, and an analysis VirusTotal has not finished when the
+poll times out is reported as a skip — never as clean.
 
 ### False-positive handling
 
--- a/docs/DEVELOPING.md
+++ b/docs/DEVELOPING.md
@@ -164,9 +164,13 @@ pnpm scan:vt -- dist/installer/installer_win.exe
 ```
 
 The publish flow also runs every built binary through VirusTotal when `VT_API_KEY` is present
-(GitHub secret on CI; root `.env` for a local `pnpm upload`), failing only when ≥
-`VT_FAIL_THRESHOLD` (default 3) engines report a binary as malicious. Without a key it just skips
-with a warning.
+(GitHub secret on CI; root `.env` for a local `pnpm upload`). The publish fails when ≥
+`VT_FAIL_THRESHOLD` (default 3) engines report a binary as malicious **or** when a veto engine
+(`VT_VETO_ENGINES`, default `Microsoft`) reports it as malicious at any count — a Microsoft/Defender
+verdict must never ship, even alone. 1–2 hits from non-veto engines warn but do not block (the
+known-FP band). The run log names the flagging engines. Without a key it just skips with a warning,
+and an analysis VirusTotal has not finished when the poll times out is reported as a skip — never as
+clean.
 
 ### False-positive handling
 
@@ -177,9 +181,11 @@ with a warning.
   <https://www.microsoft.com/en-us/wdsi/filesubmission> — select “Your app or file was incorrectly
   detected as malware” and attach the flagged binary. Microsoft can clear the hash/family in
   Defender’s cloud, which also clears it for users.
-- A durable long-term fix is **code signing** (e.g. Azure Trusted Signing); it is the only measure
-  that systematically improves AV/OS reputation, but it costs money — the measures above are the
-  zero-cost alternative.
+- A durable long-term fix is **code signing**; it is the only measure that systematically improves
+  AV/OS reputation. Paid options exist (Azure Trusted Signing), and the **SignPath Foundation**
+  sponsors free Authenticode signing for accepted open-source projects (Windows binaries only —
+  exactly the flagged artifacts here). The measures above are the zero-cost alternative while the
+  signing application is pending.
 
 ## Making changes
 
--- a/docs/DEVELOPING.md
+++ b/docs/DEVELOPING.md
@@ -156,6 +156,18 @@ pnpm scan:av -- dist/installer/installer_win.exe dist/installer/helper_win.exe
 # exit 0 = clean, exit 1 = detection, exit 2 = usage
 ```
 
+### Multi-engine check (optional — VirusTotal)
+
+```bash
+# put VT_API_KEY=... in the root .env (gitignored), then:
+pnpm scan:vt -- dist/installer/installer_win.exe
+```
+
+The publish flow also runs every built binary through VirusTotal when `VT_API_KEY` is present
+(GitHub secret on CI; root `.env` for a local `pnpm upload`), failing only when ≥
+`VT_FAIL_THRESHOLD` (default 3) engines report a binary as malicious. Without a key it just skips
+with a warning.
+
 ### False-positive handling
 
 - If a scanner flags a freshly built binary, do **not** publish it — investigate first. Local builds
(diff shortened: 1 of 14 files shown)


installer/Makefile
```diff
@@ -131,9 +142,9 @@ $(SRC_DIR)/resources.h: embed.mjs $(WEB_DIR)/index.html $(WEB_DIR)/style.css $(W
 # Windows build (cross-compile or native on Windows)
 # -mwindows: link as GUI subsystem so double-click from Explorer doesn't open a terminal
 # --verbose / --log-console flag re-attaches to parent console when run from terminal
-dist_win: resources config
+dist_win: resources config $(WIN_RES)
 	$(MKDIR)
```
onemen, line 154: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**Accepted (major, FUNCTIONAL_CORRECTNESS)** — `WIN_RES` is set only in the `ifeq ($(OS),Windows_NT)` branch, so the documented cross-compile path `make dist_win CC=x86_64-w64-mingw32-gcc` (Linux/WSL) builds `installer_win.exe` **without** the PE version resource + manifest this PR adds. CI does not cross-compile (publish-win uses a native MSYS2 win runner), so shipped artifacts are unaffected, but the documented Linux path silently omits the fix. Fix: bind the resource to the win target (`dist_win`) independent of host OS, using an overridable `WINDRES` (default `windres`; cross-compiles pass `x86_64-w64-mingw32-windres`), and update the docs cross-compile example.

onemen: Fixed in 3920158 — resolving.

onemen:
## AV / VirusTotal gate — fully green + verified end-to-end

**Latest push `7f5e60e` passes every check:**

| Gate | Result |
|---|---|
| CI (lint + format, unit tests, dev snapshot) | ✅ [`34103107177`](https://github.com/onemen/firefox-scripts/actions/runs/34103107177) |
| E2E (installer / helper / updater / portable / snap × all OS legs) | ✅ [`34103107207`](https://github.com/onemen/firefox-scripts/actions/runs/34103107207) |
| Publish gate — windows / ubuntu / mac | ✅ ×3 |

**Veto behavior verified against the real CI-built bytes** (from the dev publish run at `e4a1a28`):

- `helper_win-dev.exe` → **OK — 0 malicious / 60 undetected** (genuinely clean; continues)
- `installer_win-dev.exe` → **2 malicious — Microsoft `Trojan:Win32/Wacatac.C!ml` + Bkav** → **Microsoft veto fires → exit 1** (`pnpm scan:vt` names the engines)

The gate stops the publish **before the `Publishing` section**, so zero assets upload on a fail verdict, and the workflow's `needs:` chain halts the other platform jobs with it.

**Two gate bugs found and fixed along the way** (`f605127`, `7f5e60e`):
1. An analysis still `queued` at poll timeout was being reported as "clean" (`0/0/0/0`) — the gate now polls to completion and never reports an incomplete analysis as clean.
2. HTTP 409 on already-known bytes is handled (falls back to the file's completed stats) instead of erroring.

**Status of the underlying flag** (why the veto exists): Microsoft's cloud engine flags the current unsigned `installer_win.exe` bytes (`bb4a6126…`), so no `latest` publish should go out until this clears. In flight:
- **WDSI false-positive submission filed** — case `81e43d7f-1d0c-4798-88c2-ba05574f9034` (Medium priority)
- **SignPath Foundation application submitted** for code signing (the durable fix)

tools/scan-vt.mjs
```diff
@@ -0,0 +1,289 @@
...
+        } catch {
+          engines = undefined;
+        }
+      }
+      results.push({
+        file,
+        status: 'completed',
+        stats,
+        threshold,
+        verdict: vtVerdict(stats, threshold, engines),
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**major — Security & Privacy — VT_VETO_ENGINES never reaches the verdict.**

vtVerdict is called here without its 4th argument, so it falls back to the hardcoded VETO_ENGINES default — the VT_VETO_ENGINES env override (read by vtVetoEngines()) has no effect on the publish gate or the CLI. Fix: read the configured engines once in scanVirusTotal and pass them through:

    const vetoEngines = vtVetoEngines();
    ...
    verdict: vtVerdict(stats, threshold, engines, vetoEngines),

Without this, operators cannot extend the veto (e.g. add Bkav, the other consistently flagging engine) via the documented env knob.

tools/publish/upload.mjs
```diff
@@ -369,6 +390,22 @@ async function buildBinaries(platforms, storedHashes) {
...
   const builtInstallers = [];
   const builtHelpers = [];
+
+  // Pass 2 of the SignPath flow (--skip-build): the staged binaries ARE the
+  // final artifacts (pass 1 built them, the workflow code-signed them in
+  // place). Never rebuild or reuse — derive the built set from what exists on
+  // disk so the gates + publish below operate on the signed bytes.
+  if (SKIP_BUILD) {
+    for (const p of platforms) {
+      if (fs.existsSync(installerPath(p))) builtInstallers.push(p);
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**major — Data Integrity — --skip-build publishes an unbacked manifest when staging is empty.**

The branch derives the built set only from files present in the staging tree. If pass 1 staged nothing (or dist/.build was lost between passes), both lists stay empty: avFiles is empty so the AV and VirusTotal gates are skipped (the only pass that runs them), yet updated.installer/helper are still returned and hashes.json is published with hashes pointing at binaries never uploaded. Fail fast when this run would publish binaries without staged bytes — in dev (ALWAYS) or when installerChanged/helperChanged and the matching staged file is missing — while letting a genuinely idle prod pass (nothing changed, nothing staged) complete.

docs/DEVELOPING.md
```diff
@@ -122,6 +123,70 @@ xcode-select --install
...
+```bash
+# put VT_API_KEY=... in the root .env (gitignored), then:
+pnpm scan:vt -- dist/installer/installer_win.exe
+```
+
+The publish flow also runs every built binary through VirusTotal when `VT_API_KEY` is present
+(GitHub secret on CI; root `.env` for a local `pnpm upload`). The publish fails when ≥
+`VT_FAIL_THRESHOLD` (default 3) engines report a binary as malicious **or** when a veto engine
+(`VT_VETO_ENGINES`, default `Microsoft`) reports it as malicious at any count — a Microsoft/Defender
+verdict must never ship, even alone. 1–2 hits from non-veto engines warn but do not block (the
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**minor — Functional Correctness — non-veto guidance tied to the configured threshold.**

"1–2 hits from non-veto engines warn but do not block" is only true while VT_FAIL_THRESHOLD stays at its default of 3; the env knob is configurable, so at threshold 1 or 2 the same hits block the publish. Scope the sentence to the default ("with the default threshold of 3, 1–2 hits…") or rephrase as "hits below the configured threshold warn".

PR #148, opened 2026-09-06
Lint gate: markdownlint MD056 — merged table rows can no longer pass formatting

onemen:
## Why

PR #147 briefly shipped a **merged table row** in `docs/ci-inventory.md` — two rows glued onto one line by a careless edit:

- prettier accepts it (one syntactically valid row, just wider);
- GitHub renders it by **dropping the surplus row from the table**;
- only the `review:local` pass caught it, minutes before merge.

Nothing in the gates could see structure that prettier deliberately normalizes. This PR adds the check that can.

## What

- **`markdownlint-cli2`** (the maintained CLI over the battle-tested `markdownlint` engine — no hand-rolled tooling) added to `pnpm lint`, config in `config/.markdownlint-cli2.jsonc`.
- **MD056 `table-column-count`** is the headline rule: every row must match its header's cell count. Also on: MD055 (table pipe style), MD001 (heading increments), MD024 (duplicate headings, siblings-only), MD034 (bare URLs) — structural rules that see what prettier cannot.
- **Off**: line length, blanks-around-fences, code-fence language, first-line heading, table alignment style — prettier's domain or deliberate house style (each documented in the config with the reason).
- **Third-party skills excluded** (ADR 0022), matching the eslint markdown config's policy.
- Docs synced: AGENTS.md lint line, DEVELOPING.md checks bullet.

## Proof

Reconstructed the exact #147 broken row and ran the new gate against it:

```
broken-inventory.md:15:487 error MD056/table-column-count Table column count
[Expected: 4; Actual: 8; Too many cells, extra data will be missing]
```

`Expected: 4; Actual: 8` is the precise failure prettier could never produce. Current tree: 45 files, 0 issues.

Part of #147 follow-up

🤖 Generated with Codebuff
Co-Authored-By: Codebuff <noreply@codebuff.com>

PR #152, opened 2026-09-06
Harden #136 status table against hostile cell values; markdownlint ignores .local drafts

onemen:
Relates to #136. Follow-ups to #148; grew by two review-driven items:

## 1. Renderer hardening: `escapeTableCell` (prevention, not just detection)

`buildStatusTable` interpolates vendor-served values into the #136 meta-issue table — version strings from external release feeds, checked URLs from APIs. A raw `|` in such a value would add a cell that GitHub renders by shifting/dropping the rest of the row; a raw line break would split the row into extra markdown rows. #148 added the test-side tripwire; this makes the renderer itself defensive:

- new exported `escapeTableCell(value)`: escapes backslashes first (so a pre-escaped `\|` can never become live again), collapses `\r`/`\n` runs to a space (row-splitting vector — CodeRabbit batch finding, triaged right), then escapes pipes;
- applied to every variable cell in the row;
- **output for well-formed values is byte-identical** — the rendered issue does not change today.

Tests: hostile `155.0.1|x` version, newline-injection (`1.0\n| injected | row`), backslash-pipe combo, and `escapeTableCell` unit tests built from char codes so escaping layers can't mangle assertions.

## 2. markdownlint ignores local drafts

`markdownlint-cli2` has no gitignore support, so gitignored `.local` files errored `pnpm lint` locally. Added the same two patterns `.prettierignore` and `eslint.config.js` use — all three markdown-aware gates share one local-draft policy.

## 3. ci-inventory: deployment environments documented (Part of the #136/#123 observability thread)

New "Deployment environments" section: `release` (from #76, main-only protected-branch policy), `dev` (from 0744648, deliberately rule-free per ADR 0021), `github-pages` — plus the discovered-in-practice dev-cleanup recipe: branch deletion alone leaves the tag serving on jsDelivr; delete the release with `--cleanup-tag`.

## Validation

- 243 unit tests pass; prettier clean; eslint + C analyzer clean; markdownlint 44 files / 0 issues
- CodeRabbit batch review: 1 finding, triaged right, fixed (`44db153`), thread resolved
- CI on `e7bb6bc`: CI, E2E, URL watchdog all completed success

### DIFF
--- a/tools/check-browser-downloads.mjs
+++ b/tools/check-browser-downloads.mjs
@@ -434,14 +434,22 @@ export function formatDownloadMs(ms) {
  */
 /**
  * Escape a value interpolated into a markdown table cell: an unescaped pipe
- * would split the cell on github.com, silently shifting/dropping the cells
- * after it (the row then fails the cell-count integrity tests). Values are
- * vendor-served (version strings, checked URLs), so treat them as hostile. An
- * escaped pipe (|) renders as a literal | and does NOT split — cmark-gfm splits
- * the raw row before inline parsing, so no code-span awareness needed.
+ * would split the cell on github.com (silently shifting/dropping the cells
+ * after it), and a raw line break would split the ROW into extra markdown rows.
+ * Values are vendor-served (version strings, checked URLs), so treat them as
+ * hostile. Backslashes are escaped first so a pre-escaped backslash pipe
+ * sequence can't become live again; an escaped pipe renders as a literal pipe
+ * and does NOT split — cmark-gfm splits the raw row before inline parsing, so
+ * no code-span awareness is needed.
  */
 export function escapeTableCell(value) {
-  return String(value).replace(/\\/g, '\\\\').replace(/\|/g, '\\|');
+  // Backslashes first (a pre-escaped \| can't become live again), then line
+  // breaks (a raw \r/\n would split the ROW into extra markdown rows — same
+  // failure class as a pipe), then pipes.
+  return String(value)
+    .replace(/\\/g, '\\\\')
+    .replace(/[\r\n]+/g, ' ')
+    .replace(/\|/g, '\\|');
 }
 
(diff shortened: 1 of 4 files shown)


tools/check-browser-downloads.mjs
```diff
@@ -432,6 +432,18 @@ export function formatDownloadMs(ms) {
...
+/**
+ * Escape a value interpolated into a markdown table cell: an unescaped pipe
+ * would split the cell on github.com, silently shifting/dropping the cells
+ * after it (the row then fails the cell-count integrity tests). Values are
+ * vendor-served (version strings, checked URLs), so treat them as hostile. An
+ * escaped pipe (|) renders as a literal | and does NOT split — cmark-gfm splits
+ * the raw row before inline parsing, so no code-span awareness needed.
+ */
+export function escapeTableCell(value) {
+  return String(value).replace(/\\/g, '\\\\').replace(/\|/g, '\\|');
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**Accepted (minor, DATA_INTEGRITY_AND_INTEGRATION)** — `escapeTableCell()` escapes `\` and `|` but not `\r`/`\n`, so a hostile vendor value containing a line break would split the generated table into extra Markdown rows. Since this PR's purpose is hardening the #136 status table against hostile cell values, newline normalization belongs here. Fix: append `.replace(/[\r\n]+/g, '<br>')` and add an `assertTableIntegrity` case for a value like `'1.0\n| injected | row'`.

onemen: Fixed in 44db153: escapeTableCell now collapses 
/
 runs to a single space (backslash escaping still runs first), so a newline in a hostile vendor value can no longer split the row into extra markdown rows. Covered by a newline-injection test: row count stays 8, injected pipes render escaped.

Issue #38, opened 2026-08-24
Post-v1.0 roadmap


onemen:
Post-v1.0 roadmap

Post-v1.0 work items (milestone "Post v1.0"):

- [ ] #28 — Docs site (Astro) + user-facing docs; move gh-pages to GitHub Actions.
- [ ] #29 — Utils flavor selection (scripts / extensions / both) in installer + updater.
- [ ] #30 — Core code test coverage (stub smoke tests + Nightly leg + test-required convention).
- [ ] #31 — Browser-matrix expansion: Waterfox / Zen / Firefox Nightly CI legs.
- [ ] #32 — UAC/admin-rights E2E automation (self-hosted runners).
- [ ] #33 — Publish pipeline automation: build-and-upload.yml, deterministic-publish check, helper sha256 trust.
- [ ] #34 — Installer UI polish + updater UX follow-ups.

Portable browser expansion is intentionally deferred until Firefox Release portable coverage is stable in the v1.0 E2E gate (#56).

- [ ] Add portable-layout E2E legs for forks (LibreWolf, Floorp, Zen, Waterfox, and others as their official distribution formats allow).
- [ ] Add fork-specific portable download/extraction recipes without weakening the required Firefox Release legs.
- [ ] Revisit whether fork portable legs should remain advisory or become required after reliable CI coverage exists.

## CI safety-net follow-ups (post-v1.0)

[...]

PR #166, opened 2026-09-08
feat(installer): ARM64 Linux builds — installer_linux_aarch64 + helper twin

onemen:
Linux half of the pre-1.0 arch-coverage scope agreed in #4 (macOS universal is #165 / #132). Firefox ships arm64 Linux builds, so the installer and helper must run there too.

## Design

Separate **assets built from identical sources** — Linux has no fat-binary convention, and the hash manifest covers the **source tree**, not binaries (ADR 0002/0008), so the manifest format, hash logic, and C/JS parity stay untouched. New assets: `installer_linux_aarch64[-dev]` (release asset) + `helper_linux_aarch64[-dev]` (Pages-only, like the other helpers).

## What

- **`installer/Makefile`** — `dist_linux_aarch64` + `helper_linux_aarch64` targets; `AARCH64_CC` override (default `aarch64-linux-gnu-gcc`; a native arm64 host passes `AARCH64_CC=cc`).
- **`installer/src/platform.h`** — `__aarch64__` builds self-update from `installer_linux_aarch64` (an x86_64 binary cannot run on the host).
- **`updater.js`** (updater-ui package) — `installerFilename()`/`helperFilename()` are arch-qualified: a native arm64 Firefox (aarch64 in `XPCOMABI`/`oscpu`) downloads `helper_linux_aarch64` for elevated copies and points the manual panel at the arm64 installer. The aarch64 helper must match the running arch — an x86_64 helper can't exec on arm64 kernels without the 32-bit compat layer.
- **`upload.mjs`** — new `aarch64` platform entry; `--platform=linux` **implies the aarch64 twin** (same job, one toolchain install), `--platform=aarch64` builds it alone; `AARCH64_CC` passes through to make.
- **CI** — `ci.yml` publish-gate Linux leg and `pages.yml` `publish-linux` install `gcc-aarch64-linux-gnu`; `docs/ci-inventory.md` rows updated (the AGENTS.md sync rule).
- **Docs** — README download list, DEVELOPING.md build + release-asset sections, status-logic.md, auto-updater.md.
- **Installer test harnesses** — `test_hash.mjs` / `test_self_update.mjs` pick the binary matching `process.arch` so an arm64 Linux machine can run them natively.

## Validation

- `make -n dist_linux_aarch64` / `helper_linux_aarch64`: correct compile lines with `aarch64-linux-gnu-gcc`.
[...]

PR #165, opened 2026-09-08
feat(installer): universal macOS build (x86_64 + arm64) with CI slice check

onemen:
Fixes #132 (macOS universal installer — the decision, implemented: one universal `installer_mac` asset).

## What

`dist_mac` currently builds a native-arch-only installer, so an Apple Silicon (or Intel) user can get an installer that cannot run. `helper_mac` is already universal (`-arch x86_64 -arch arm64`); this PR mirrors that for the installer:

- **`installer/Makefile`** — `dist_mac` now compiles `-arch x86_64 -arch arm64` into one `installer_mac` (and `installer_mac-dev` in dev mode).
- **CI verification** — the e2e macOS leg (which already builds the dev installer) now asserts the binary is really universal: `lipo -archs` must contain both slices (hard error otherwise), then `lipo -thin` extracts each slice and **executes it** (`--version`) — proving each architecture actually runs, not just that the fat binary exists.
- **`docs/DEVELOPING.md`** — the macOS build section documents the universal build and the `lipo` check.

## Why one asset (ADR 0019 interplay)

Asset names stay **stable and unversioned**: `installer_mac` keeps its name, so the updater (`updater.js` self-update URL), the release-asset list, and ADR 0019's fixed-name rule are untouched — no per-arch asset names to maintain. Build cost on CI is one extra codegen slice per compile (the same cost `helper_mac` already pays).

## Validation

- `make -C installer -n dist_mac` shows the `-arch x86_64 -arch arm64` flags in the link command.
- `pnpm lint` / `pnpm format` clean (includes gcc -fanalyzer over the unchanged C sources).
- The slice-execution check itself can only run on a macOS runner (e2e.yml macos leg); local verification here is the dry-run + the identical-flag mirror of the proven `helper_mac` recipe.

🤖 Generated with Codebuff
Co-authored-by: Codebuff <noreply@codebuff.com>

PR #164, opened 2026-09-08
feat(publish): abort prod publish when env overrides redirect the target

onemen:
Part of #33 (staging-target guard item; the other #33 items stay post-v1.0).

## What

`tools/publish/paths.js` gives environment variables precedence over `config/installer.conf` (env-over-conf), so a stray `REPO_OWNER=...` / `HASHES_URL=...` in the shell or CI environment silently redirects a prod publish to an unintended release, Pages branch, or manifest host.

This PR makes that loud instead of silent:

- **New `tools/publish/stagingGuard.mjs`**, wired into `upload.mjs` right after flag parsing — before anything is built or uploaded.
- **Prod + override → loud STAGING banner + hard abort** (exit before any build step). Baseline is `installer.conf` itself: even a value *equal* to the conf value is reported — publishing must not depend on ambient shell state.
- **Dev mode → warn-only**, and only for repo-identity keys (`REPO_OWNER`/`REPO_NAME`/`ZIP_PAGES_REPO`) — dev artifacts are disposable and live in the `dev-build-<id>` namespace anyway.
- **`--local` snapshots are exempt** (they touch no GitHub target).
- **Escape hatch** for an intentional staging rehearsal: `FIREFOX_SCRIPTS_ALLOW_STAGING=1` prints the banner and continues.
- **`.env-example`** documents the full key list, the abort behavior, and the escape hatch (the `#33` checklist item "Document the staging keys in .env-example (same key set, with prod-safety warnings)").

Covered keys: `REPO_OWNER`, `REPO_NAME`, `RELEASE_NAME`, `ZIP_PAGES_REPO`, `ZIP_PAGES_BRANCH`, `HASHES_URL`, `ZIP_BASE_URL`, `ZIP_PAGES_URL`, `HELPER_BASE_URL`, `UI_BASE_URL`.

## Validation

- 12 new unit tests in `test/unit/publish/stagingGuard.test.mjs` (conf parsing, per-mode key scoping, baseline-equal overrides, banner rendering, all four guard behaviors); full suite 292/0.
- Live probe: `REPO_OWNER=someone-else node tools/publish/upload.mjs --mode=prod` → banner + `Prod publish aborted`, exit 1, zero build steps.
- `pnpm lint` / `pnpm format` clean; verified CI sets none of these keys today (no false-positive risk in workflows).

🤖 Generated with Codebuff
Co-authored-by: Codebuff <noreply@codebuff.com>

PR #168, opened 2026-09-09
fix(core): no-op guard for BootstrapLoader.js top-level patches on Waterfox

onemen:
Fixes part of #38 (pre-1.0 scope: Waterfox BootstrapLoader double-load guard)

## Problem

`config.js` already skips loading `BootstrapLoader.js` on Waterfox via a name regex (`/waterfox/i`), because Waterfox bundles its own legacy-extension BootstrapLoader. But that protection lives entirely in config.js — a **user-modified `config.js`** that loads `BootstrapLoader.js` unconditionally would run every top-level side effect a second time on Waterfox:

- duplicate `chrome-document-loaded` observers (updater init + about:addons patching)
- `addon-card` / `addon-options` prototypes patched twice (chained wrappers)
- second `AddonManager.addExternalExtensionLoader(BootstrapLoader)` registration
- `XPIDatabase.isDisabledLegacy` / `XPIExports.verifyBundleSignedState` re-patched

## Change

Guard **all top-level side-effect blocks** in `BootstrapLoader.js` with a `waterfoxBrand` flag so the whole file becomes a no-op on Waterfox, regardless of who loads it:

- updater-init observer
- about:addons prototype patches
- XPIDatabase / XPIExports patches
- `addExternalExtensionLoader` + the isReady reload loop

The constant is named `waterfoxBrand` — deliberately **not** `isWaterfox`, since both files evaluate into the same `loadSubScript` global lexical environment and a redeclaration would throw on every browser.

Behavior on Waterfox is otherwise unchanged: `userChrome.js` and the updater still load normally, and `userChrome.js` registers its own updater-init observer (`scriptsUpdater.sys.mjs` is idempotent).

## Tests

New `test/unit/core/bootstrapLoader.test.mjs` (6 tests) evaluates the full file in a Node vm with mocked Firefox globals:

- Waterfox (incl. name variants): zero observers, zero loader registrations, XPIDatabase/XPIExports untouched
- Firefox: both observers, loader registered once, patches applied
- isReady reload loop only reloads enabled legacy add-ons
- **Shared-scope regression test**: `config.js` + unconditionally-loaded `BootstrapLoader.js` in one vm context — no lexical redeclaration crash, still a no-op on Waterfox

`pnpm test` 286/0 fail · `pnpm lint` clean · `pnpm format` clean.

🤖 Generated with Codebuff

### DIFF
--- a/core/chrome/utils/BootstrapLoader.js
+++ b/core/chrome/utils/BootstrapLoader.js
@@ -6,16 +6,6 @@
 
 const Services = globalThis.Services;
 
-// Waterfox bundles its own legacy-extension BootstrapLoader, so our copy must
-// never run there — config.js skips loading this file on Waterfox. If a
-// user-modified config.js loads it unconditionally anyway, the top-level
-// patches below would double-register observers, double-patch the add-ons UI
-// and register the external loader a second time. Guard them all with this
-// flag so the whole file stays a no-op on Waterfox (userChrome.js and the
-// updater still load normally there — userChrome.js registers its own
-// updater-init observer).
-const waterfoxBrand = /waterfox/i.test(Services.appinfo.name);
-
 ChromeUtils.defineESModuleGetters(this, {
   Blocklist: 'resource://gre/modules/Blocklist.sys.mjs',
   ConsoleAPI: 'resource://gre/modules/Console.sys.mjs',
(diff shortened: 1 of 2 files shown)


onemen (commented):
🤖 AI review triage (Codebuff agent — result of the ADR 0020 local review step, `pnpm review:local`)

Reviewed `main...HEAD` (2 files: `core/chrome/utils/BootstrapLoader.js`, `test/unit/core/bootstrapLoader.test.mjs`). **0 findings** — nothing to triage, no threads opened.

Notes from the run:
- The guard covers every top-level side-effect block in the file (verified by enumeration, not just the diff: updater-init observer, about:addons prototype patches, XPIDatabase/XPIExports patches, `addExternalExtensionLoader` + isReady reload loop).
- The `waterfoxBrand` name avoids the loadSubScript shared-lexical-scope collision with `config.js`'s `isWaterfox`; the shared-scope vm test locks this in.

onemen:
🤖 Answering the three review questions (changes pushed in f661468):

**1. One function, one check — done.** All top-level side effects now live in `initBootstrapLoader()`; the decision happens once via `bootstrapLoaderBundled()` at the bottom of the file. This also structurally answers (2): nothing inside the init function can leak global bindings on any browser.

**2. Global bindings — verified clean.** The only top-level bindings are the pre-existing ones (Services, the ES-module getters, the `BootstrapLoader` object, detection/init functions and their constants — same as before this PR, minus `orig_verifyBundleSignedState`, which is now init-scoped). No new globals added.

**3. Yes — via the AddonManager external-loader registry.** Gecko exposes a public `externalExtensionLoaders` Map getter (AddonManager.sys.mjs, [getter at line 4002](https://searchfox.org/firefox-main/source/toolkit/mozapps/extensions/AddonManager.sys.mjs#4002)), and `addExternalExtensionLoader` registers keyed by `loader.name` (line 2757). `bootstrapLoaderBundled()` now checks:
- brand regex (primary — autoconfig runs before AddonManager startup, so the registry may not be populated yet on Waterfox itself; deterministic),
- registry: `'bootstrap'` key present, or any loader with `manifestFile === 'install.rdf'` (secondary — catches a rebranded fork whose bundled loader is already registered).

Bonus property of the registry check: a second evaluation of this file in a fresh scope is inert on **any** browser, because our own first registration is then visible. New vm tests cover both the rebrand case and the double-evaluation case.

onemen:
🤖 AI review triage (Codebuff agent — result of the final `review:batch` CodeRabbit pass over all four open PRs: #164, #165, #166, #168)

One batched review over the combined 21-file diff (`tools/ci/batch-review.mjs`). **0 findings on this PR** — nothing to triage, no threads opened. Combined with the earlier ADR 0020 local reviews and the green CI runs, this PR is ready for merge.

onemen:
🤖 Correction to [my earlier comment](https://github.com/onemen/firefox-scripts/pull/168#issuecomment-5598092975) — it contained a factual error, caught by empirical verification in PR review:

**`AddonManager.externalExtensionLoaders` is NOT a thing.** The `externalExtensionLoaders` getter lives on the **`AddonManagerPrivate`** export of `AddonManager.sys.mjs` (line 4002; the Map itself in `AddonManagerInternal`, line 565; consumers: `XPIProvider`/`XPIInstall` via `AddonManagerPrivate`). The public `AddonManager` object never exposes it — so the previous check read `undefined` on every browser and silently degraded to the brand-regex-only guard (no crash, thanks to the `try`/fallback, but the behavioral signal never worked).

Fixed in d6620d2: `bootstrapLoaderBundled()` now imports and reads `AddonManagerPrivate.externalExtensionLoaders` — the same surface Gecko's own XPIProvider uses. The tests now model the real surface (dynamic getter re-reading the live Map) instead of the invented one, so a regression to the wrong object would fail them. All 9 vm tests pass, suite 289/0, lint/format clean.

Lesson recorded: searchfox line numbers alone don't tell you *which object* in the module carries a property — verify the enclosing scope, or empirically, before claiming an API.

Issue #102, opened 2026-09-03
Lack of ui folder in the utils.zip


Merci-chao:
@onemen I've found that the `utils.zip` in the [release page](https://github.com/onemen/firefox-scripts/releases), which for manual installation, lacks of the `ui` folder. Is it normal?
<img width="522" height="288" alt="Image" src="https://github.com/user-attachments/assets/ce319030-5c05-4f33-a251-70a1ee106e78" />

onemen:
Yes this is planed

in order to eliminate the need to update utils.zip when UI change the file `https://github.com/onemen/firefox-scripts/blob/gh-pages/updater-ui.zip` automatically download when `utils.zip` need update if it is missing
this file also installed by the installer

I am currently working on the tests to verify that the workflow is working for all operating system and Firefox and forks version

I sold also work if you manually install the new utils.zip
to test it after you install it in a profile, open one of the files in chrome/utils and add comments

start Firefox, the updater UI should be open in new tab

Merci-chao:
I have tried in a new clean profile, and just extract the `utils.zip` into the `chrome` folder, the following error shows up in the console:
<img width="1077" height="83" alt="Image" src="https://github.com/user-attachments/assets/3c562241-043f-439d-a6f0-251b585d8352" />

I have added a comment in the userChrome.js and restart, still nothing happens, no updater UI:
<img width="630" height="528" alt="Image" src="https://github.com/user-attachments/assets/83419d2e-8e84-47d5-a8f0-07239524734b" />

And there are no related preferences show in about:config:
<img width="1074" height="242" alt="Image" src="https://github.com/user-attachments/assets/d11087bb-41a3-4db8-a48d-b61ad9be5f4e" />

Merci-chao:
Things only happen after I copy the `ui` folder into the updater folder.

Merci-chao:
Or maybe the `updater-ui.zip` should also be mentioned in the manual installation section of README?

onemen:
@Merci-chao 

you do not need to manually copy the ui
remove it.
then install only the utils.zip normally

Merci-chao:
Then I'm confused. 
I downloaded the `utils.zip` from the releases page, set it up normally, modified the `userChrome.js` a bit, restarted Firefox and nothing happened.
No updater page nothing.

onemen:
@Merci-chao 

see comments in chrome/utils/updater/scriptsUpdater.sys.mjs 

clear the prefs
[...]

Issue #72, opened 2026-08-26
Publish pipeline: date-stamped component releases (scripts-<date> / installer-<date>) with make_latest=false, per ADR 0019


onemen:
Part of the v1.0 release gate ([#4](https://github.com/onemen/firefox-scripts/issues/4)); related to publish automation ([#33](https://github.com/onemen/firefox-scripts/issues/33) umbrella).

Decision recorded in [ADR 0019](./docs/decisions/0019-release-versioning.md):

- Artifact names stay permanent and unversioned (`utils.zip`, `fx-folder.zip`, `updater-ui.zip`, `installer_<os>`, `helper_<os>`).
- New per-component, date-stamped releases: `scripts-<YYYY-MM-DD>` (the zips) and `installer-<YYYY-MM-DD>` (installer + helper binaries), created **only when that component changed** and with `make_latest=false` so GitHub keeps `latest` as the "Latest" release.
- `latest` keeps the **complete asset set** — the updater and README/docs continue to point at `releases/download/latest/...` unchanged.

**Status 2026-09-11: implemented by #176 (merged 2026-09-09).** One deviation from the ADR wording as written here: the REST API has no `make_latest` parameter, so component releases are created with `prerelease: true` — prereleases can never take GitHub's "Latest" badge, which preserves the invariant (documented in `tools/publish/componentReleases.mjs` and #4).

Pipeline changes this implies (in `tools/publish/upload.mjs` + `uploadUtilsZip.mjs`) — all done in #176:

- [x] `getOrCreateRelease`/upload flow creates/updates the two component releases alongside `latest`, skipping a component whose assets are unchanged (reuse logic already exists per-asset in `upload.mjs`) — buckets sync only when that component was rebuilt; same-day republish replaces assets in place.
- [x] Component releases carry the date (UTC) and a body describing what changed — body lists the artifacts and points back at `latest`/gh-pages; rewritten in place on same-day republish.
- [x] Verify the updater/installer keep working with `latest` only (no code change expected — URLs are fixed to `latest`) — #176 touched no core files; live end-to-end confirmation lands with the RC rehearsal tracked in #4.
- [x] `devClean.mjs` / dev mode untouched (dev stays `dev-build-<id>` only) — the sync is gated `PUBLISH_MODE === 'prod'` in `upload.mjs`.

onemen:
Closing as implemented — status for the record:

[...]

PR #176, opened 2026-09-09
feat(publish): date-stamped component releases alongside latest (#72)

onemen:
Implements #72 · ADR 0019's component-release decision · Pre-1.0 item 4 of #4.

## What

Every **prod** publish now also syncs two date-stamped component releases:

| Tag | Carries | When |
|---|---|---|
| `scripts-<YYYY-MM-DD>` | rebuilt package zips (`utils.zip`, `fx-folder.zip`) | any zip rebuilt |
| `installer-<YYYY-MM-DD>` | rebuilt installers + helpers (+ `.sha256` sidecars from #174) | any installer/helper rebuilt |

## The mechanism decision (why prerelease=true)

ADR 0019 says the date tags exist with `make_latest=false` — but the REST API has **no** `make_latest` parameter, and GitHub's "Latest" badge follows the newest *created* non-prerelease, non-draft release. A freshly created date tag would therefore steal the badge from `latest` on every publish. The only per-release REST control is `prerelease: true` — prereleases are **never** "Latest". So: component releases are created as prereleases. This is also semantically honest (a frozen per-component snapshot, not "the" release), removes all create-order coupling, and needs no post-hoc re-pin. Documented in the module header.

## Behavior

- Buckets sync **only when something in them was rebuilt** — an idle run leaves the existing tags frozen at their own dates.
- Same-day republishes replace assets and rewrite the body in place (idempotent).
- Bodies list the artifacts and point humans back at `latest`/gh-pages; `hashes.json` stays the machine source of truth.
- `updater-ui` stays Pages-only (never a release asset, per #102).
- The whole sync is **fail-soft**: a component-release failure warns, never fails the publish (`latest` + Pages are the contract).
- Dev mode unchanged (dev-build releases are already prereleases).

## Validation

- `pnpm test` — 317 pass / 0 fail (5 new tests: date/tags/grouping/body rendering)
- `pnpm lint` ✓ · `pnpm format` ✓
- Live verification lands with the next real prod publish (the RC rehearsal is #4's last item)

🤖 Generated with Codebuff


tools/publish/componentReleases.mjs
```diff
@@ -0,0 +1,174 @@
...
+      const {helperSha256Sidecar} = await import('./hashUtils.mjs');
+      const {readFileSync} = await import('fs');
+      for (const p of installer) {
+        const inst = installerAssetName(p);
+        const helper = helperAssetName(p);
+        const sidecar = helperShaAssetName(p);
+        assets.set(inst, installerPath(p));
+        assets.set(helper, helperPath(p));
+        // Sidecar regenerated from the staged bytes (issue #33 contract).
+        assets.set(sidecar, helperSha256Sidecar(readFileSync(helperPath(p)), helper));
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run over #163/#169/#174/#175/#176)

**Accepted — correct Major.** `groupBuilt` unions installer+helper platforms, but this loop read both staged paths for every unioned platform. A partial rebuild (e.g. helper leg skipped) made `readFileSync(helperPath(p))` throw ENOENT inside the fail-soft wrapper — silently dropping the whole `installer-<date>` release even though its installer was built.

Fixed in 32afdbd: extracted `componentAssets()` (pure, injectable accessors) — each artifact is guarded on its own built list, and an empty asset map skips the release instead of creating a bodyless one. Covered by two new unit tests including this exact partial-rebuild case.

PR #182, opened 2026-09-12
E2E speedup: unit tests for hash logic + local manifest server

onemen:
## Summary

- **26 new unit tests** for the hash/decision logic extracted from `scriptsUpdater.sys.mjs` — pure Node reimplementation of `computeFilesHash`, `packageNeedsUpdate`, manifest parsing. Runs in ~2.5s, no Firefox needed.
- **Local manifest server** (`test/e2e/shared/localManifestServer.mjs`) — tiny HTTP server that serves a pre-baked `hashes.json` for "no update" E2E scenarios. The scheduler fetches from localhost (~1ms) instead of the network, eliminating latency and CI variance.
- **Faster no-tab scenarios** — `runNoTabScenario` (Scenarios 4, 5) and `manual-install` Phase 1 now use the local server with `HASHES_URL` override. Blind wait cut from 3s to 500ms.
- **Timing instrumentation** — no-tab scenarios log `seed`, `serverStart`, `launch`, `waitBrowser`, `total` durations for CI performance tracking.

## Context

The E2E matrix already runs on every PR as a required check. The "no update" scenarios (tab should NOT open) were waiting 3s blind + network fetch time for the scheduler to decide. Since we control the test fixture, we know the hashes match — the network fetch is unnecessary.

The unit tests cover the pure-logic parts of the updater that are extractable to Node without a Firefox context. They don't replace the E2E scenarios but catch regressions in the hash algorithm and update decision logic quickly.

## Validation

- `pnpm lint` — passes
- `pnpm format` — passes  
- `pnpm test` — 346 tests pass (3 platform-specific Linux tests skipped)
- Pre-push gate (lint + format + test) — passes

## Expected CI impact

Per no-tab scenario: ~2.5s savings (3s margin → 500ms). Across the Firefox matrix (9 legs × 2-3 no-tab scenarios each): measurable but not revolutionary. The real win is eliminating network variance from these scenarios — they become deterministic.

Tab-open scenarios (1, 2, 3, 6, 7-phase2, 8) are unchanged — they need the real manifest to prove the tab opens for the right reason.

🤖 Generated with Codebuff


onemen:
🤖 AI review triage (Codebuff agent — review + CI failure diagnosis of this PR)

## Verdict: worth keeping — fixed, all checks green

The idea is sound: 26 pure-Node unit tests for the hash/decision logic, and a localhost manifest server that makes the "no update" scenarios deterministic. But the PR shipped a regression that failed `manual-install-upgrade` on **all 15 E2E legs**, plus ~140 lines of speculative diagnostics and two lint issues.

## Root cause of the CI red

Phase 1 of `manual-install-upgrade` pointed `HASHES_URL` at the local manifest server via `extraPrefsFirefox`. Puppeteer writes that pref into the profile's `prefs.js` at launch, and **Firefox flushes it back on close**. Phase 2 relaunches on the *same profile* — so it inherited the override pointing at the now-dead `http://127.0.0.1:<port>/hashes.json`. The manifest fetch failed, `checkScriptsUpdateNeeded` exited silently as "no update" (by design for network failures), and the updater never activated: no tab, no `ensureUpdaterUi` extraction.

The PR's own `[diag:phase2]` logs proved it: the override was cleared from the JS prefs object, but "clearing the object" cannot remove a pref already flushed to disk. `manual-install` is the only scenario that relaunches the same profile — scenarios 4/5 (which keep the local server, correctly) always use fresh profiles.

## What changed in the fix (2 commits)

1. Merge of main (up to date with the latest tip).
2. `fix(e2e): phase-2 relaunch poisoned by phase-1 HASHES_URL override pref`:
   - Phase 1 restored to its pre-PR form (3s negative-assertion margin, no manifest server) with a comment explaining the cross-phase pref hazard.
   - The ~140 lines of `[diag:phase2]` debug output removed — the PR's real wins (scenarios 4/5 timing + unit tests) are untouched.
   - `localManifestServer.mjs` prettier-formatted and three dead vars dropped. Note: the file is invisible to plain `pnpm lint` via the global `**/*local*.*` eslint ignore, which is why the PR's original lint run never saw its 8 violations.

## Result

- All 15 updater E2E legs green (was 15/15 red); full board green including CI/E2E gates.
- Scenario 4/5 still get the PR's intended win: `total=6.3s` with `postWait=500ms` (was 3s+ blind wait).
- One transient flake along the way (`Install firefox-dev` failed because the cached installer exe was still locked by the cache-restore tar — Windows file-lock race, not this PR; passed on `gh run rerun --failed`).

test/e2e/shared/localManifestServer.mjs
```diff
@@ -0,0 +1,242 @@
...
+  // Extract utils to a temp dir to compute the matching hash
+  const staging = fs.mkdtempSync(path.join(REPO_ROOT, 'dist', 'fxs-manifest-server-'));
+  let manifest = manifestOverride;
+  if (!manifest) {
+    extractZip(utilsZip, staging);
+    manifest = buildMatchingManifest(staging, snapshotDir);
+  }
+  const body = JSON.stringify(manifest, null, 2);
+
+  const port = await findFreePort();
```
onemen: 🤖 **AI review — agent-posted (ADR 0020) — result of the CodeRabbit review:batch run** — major — `findFreePort()` races the real server: the probe socket binds `:0`, reads the assigned port, **closes the socket**, and only then does the manifest server `listen(port)`. In that window another process can claim the port (CI runners are exactly such an environment), and the listen then fails with EADDRINUSE, failing the scenario. Accepted: bind the real server with `listen(0)` directly and read the OS-assigned port from `server.address().port` after the bind resolves. Fix will land as a follow-up commit on this branch.

test/unit/e2e/scriptsUpdater-hash.test.mjs
```diff
@@ -0,0 +1,340 @@
...
+  const sorted = [...files].sort((a, b) => a.localeCompare(b));
+  const hash = crypto.createHash('sha256');
+  for (const relative of sorted) {
+    hash.update(relative + '\n');
+    const abs = path.join(baseDir, ...relative.split('/'));
+    if (fs.existsSync(abs) && fs.statSync(abs).isFile()) {
+      hash.update(fs.readFileSync(abs));
+    }
+  }
+  return hash.digest('hex');
```
onemen: 🤖 **AI review — agent-posted (ADR 0020) — result of the CodeRabbit review:batch run** — major — The test duplicates `computeFilesHash`/decision logic as local reimplementations, so it can never catch a regression in the real `scriptsUpdater.sys.mjs` — a bug in production passes as long as the copy is also wrong. Accepted with an adjustment: `scriptsUpdater.sys.mjs` cannot be imported directly in Node (`ChromeUtils is not defined` at module top), and the C-parity contract is already covered separately by `installer/test/test_hash.mjs` + `pnpm test:hash`. So: drop the copy-pasted `computeFilesHash` and instead evaluate the **real** `.sys.mjs` in a `vm` sandbox with stubbed Firefox services — the exact pattern `test/unit/core/scriptsUpdater-channel.test.mjs` established for this same module (#189, merged today). The decision-logic tests (`packageNeedsUpdate` etc.) keep a *reference* implementation as the expected value; the assertions compare production vs reference, which is a genuine tautology-breaker instead of two copies of the same code. Fix will land as a follow-up commit.

PR #185, opened 2026-09-12
publish: Latest Scripts scheme — full component releases, make_latest pin, helpers off releases

onemen:
Part of #72. Stacked on #184 (shares `componentReleases.mjs`). Implements the scheme designed in the release-page mock review and summarized in [this #4 comment](https://github.com/onemen/firefox-scripts/issues/4#issuecomment-5645375392).

## What

1. **Full component releases + badge pin.** `scripts-<date>` / `installer-<date>` are created as full releases (no prerelease flag); after every publish, `pinLatestRelease()` re-pins the Latest badge onto `latest` via `make_latest=true` (Update-a-release). GitHub renders the badge-holding release as the hero card at the top of the page — so the page reads: **Latest Scripts** hero first, frozen date tags below. Replaces the old `prerelease=true` badge-guard (whose premise — "REST has no `make_latest`" — was incorrect; verified in octokit's generated types).
2. **Helpers leave the release page.** `groupBuilt`'s installer bucket is installer binaries only; `componentAssets` no longer adds helper binaries or `.sha256` sidecars. Helpers stay gh-pages-only per ADR 0019's original rule; a helper-only rebuild creates **no** release tag. The maintainer's call from the mock review (they are not user downloads — the updater fetches them with verify-before-execute).
3. **Per-file dates, twice.** Release bodies list each asset with `— updated <date>` (bodies are always visible; asset rows are collapsible on GitHub), and assets upload with a `Updated <date>` **label** (≤50 chars) as the bonus. Dates come from the merged manifest's per-package source-commit date — the same date the updater tab shows. (GitHub renders the label in place of the filename on the asset row; full name on hover.)
4. **ADR 0019 amended in place** (badge-mechanics paragraph); the stale "REST has no make_latest" note on #4's #72 entry gets corrected when this merges.

## Out of scope (sequenced)

Zip entry timestamps (files *inside* the zips showing the release date) — separate small PR, needs a date-source decision. `latest`'s own body/date list stays as is here.

## Validation

- `node --test test/unit/publish/componentReleases.test.mjs` ✓ 7/7 (rewritten contracts: helpers never join, per-file dates, `Updated` labels)
- `pnpm lint` ✓ · `pnpm format` ✓ · `pnpm test` ✓ 326/0
- `check:decisions` ✓ (0019 amendment keeps links/statuses valid)

🤖 Generated with Codebuff

### DIFF
--- a/tools/publish/upload.mjs
+++ b/tools/publish/upload.mjs
@@ -89,7 +89,7 @@ import {
   REPO_ROOT,
 } from './publishCommon.mjs';
 import {pagesIndex, uploadFilesToPages} from './uploadToPages.mjs';
-import {syncComponentReleases} from './componentReleases.mjs';
+import {pinLatestRelease, syncComponentReleases} from './componentReleases.mjs';
 import {scanBinaries} from '../scan-av.mjs';
 import {scanVirusTotal} from '../scan-vt.mjs';
 import {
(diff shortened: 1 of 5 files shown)


tools/publish/upload.mjs
```diff
@@ -687,18 +687,21 @@ async function publishToGitHub({
...
+  // (helpers are gh-pages-only — never release assets). Full releases, then the
+  // Latest badge is re-pinned onto `latest` via make_latest (the Latest Scripts
+  // scheme — the badge release renders as the page's hero card). Skipped on
+  // idle runs (nothing rebuilt → tags stay frozen).
   if (PUBLISH_MODE === 'prod' && anythingUploaded) {
     await syncComponentReleases(octokit, {
       builtZips,
       builtInstallers,
       builtHelpers,
+      manifest: merged,
```
onemen, line 700: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run) — major — **Right, fixed in 0ce0dd4.** `upload.mjs` passed shorthand `{merged}` while `syncComponentReleases` destructures `manifest`, so the parameter arrived as `undefined`: the per-package date lookup threw inside the fail-soft try/catch and every release silently lost its "Updated YYYY-MM-DD" labels. Fixed by passing `manifest: merged` at this call site; a unit assertion now guards the destructuring contract.

PR #190, opened 2026-09-12
e2e: never let test browsers register Windows startup entries

### LINKED ISSUE #191: E2E test installs register Windows startup entries (HKCU Run debris)
E2E test installs register Windows startup entries (HKCU Run debris)

## Symptoms

30+ `Mozilla-Firefox-<hash> = <firefox.exe> -os-autostart` entries accumulated under `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, all pointing at throwaway test installs (`%TEMP%\fxs-e2e-*`, `%TEMP%\fxs-smoke-*`, repo-local portables). Windows 11's Settings → Apps → Startup lists one row per Run value, so the page filled with dead Firefox rows.

## Root cause (corrected)

The Run values are written by Firefox's **launch-on-login auto-enable**, not by `RegisterApplicationRestart`:

- `DefaultLaunchOnLogin.maybeEnableOnFirstRun` (browser-idle-startup) runs on the **first run of a fresh profile** of an official build; when the Nimbus pref `browser.startup.windowsLaunchOnLogin.defaultEnabled` is set, it calls `LaunchOnLogin.enable()` → `createLaunchOnLoginRegistryKey`, which writes `"Mozilla-Firefox-<installHash>" = "\"<XREExeF>\" -os-autostart"` (see `WindowsLaunchOnLogin.sys.mjs`; the value name comes from `vendor-name-<getInstallHash()>`, so every distinct install directory gets its own entry).
- Every E2E leg launches a **fresh install in %TEMP% + a fresh profile** → `profileService.isFirstRun` is true on every run → every run of an official build could register another startup entry.
- Firefox source searchfox refs: `browser/components/defaultlaunchonlogin/DefaultLaunchOnLogin.sys.mjs`, `toolkit/modules/launchonlogin/WindowsLaunchOnLogin.sys.mjs`.

`toolkit.winRegisterApplicationRestart` (the Restart Manager registration, `nsAppRunner.cpp`) is a separate mechanism and does not produce these Run values; it is disabled in the harness only as a belt-and-suspenders measure.

## Why day-to-day puppeteer use (e.g. firefox-updater / Tab Mix Plus development) never hit this

Those launches reuse **persistent, already-initialized profiles**, so first-run never fires and the auto-enable never runs. Only fresh-profile test installs are exposed.

## Fix direction

- `launchFirefox` (test/e2e/shared/helpers.mjs) launches every test browser with `browser.startup.windowsLaunchOnLogin.defaultEnabled=false` and `browser.startup.windowsLaunchOnLogin.alreadyApplied=true` (plus `toolkit.winRegisterApplicationRestart=false` as belt).
- No registry sweeping: the harness never mutates the user's Run key — an entry there may be intentional. Prevention only.

PR: #190 (contains the harness fix; the dev machine's 30 debris entries were already removed by hand).

onemen:
## Problem

Every run of the E2E/smoke suite left a new entry in the user's Windows **Startup apps** page: HKCU `Run` values named `Mozilla-Firefox-<installHash>` pointing at the throwaway `%TEMP%` test install with `-os-autostart`. 30 had accumulated on the dev machine.

## Root cause (verified against Firefox source)

The Run values are written by Firefox's **launch-on-login auto-enable**, not by `RegisterApplicationRestart`: on the **first run of a fresh profile** of an official build, `DefaultLaunchOnLogin.maybeEnableOnFirstRun` consults the Nimbus pref `browser.startup.windowsLaunchOnLogin.defaultEnabled` and calls `LaunchOnLogin.enable()` → `createLaunchOnLoginRegistryKey` (`WindowsLaunchOnLogin.sys.mjs`), writing `"Mozilla-Firefox-<installHash>" = "\"<exe>\" -os-autostart"`.

Every E2E leg launches a fresh install + fresh profile → `isFirstRun` true every time → one new entry per install directory per run. This is also why day-to-day puppeteer use against **persistent profiles** (e.g. the Tab Mix Plus updater tooling) never registers anything: first-run never fires there.

## Fix

- `STARTUP_HYGIENE_PREFS` (test/e2e/shared/helpers.mjs) carries the three kill switches:
  - `browser.startup.windowsLaunchOnLogin.defaultEnabled=false` — the Nimbus pref the auto-enable consults (user-pref overrides experiment values)
  - `browser.startup.windowsLaunchOnLogin.alreadyApplied=true` — skips the auto-enable path entirely
  - `toolkit.winRegisterApplicationRestart=false` — belt: the separate Restart Manager registration
- **puppeteer path:** `launchFirefox` injects the prefs via `extraPrefsFirefox` (puppeteer overwrites a caller-written user.js, so this is the only channel there).
- **detached-spawn path:** `launchDetachedFirefox` (installer E2E's restart-scope leg — bystander A + target B) now seeds the prefs into the fresh profile's user.js via `seedStartupHygienePrefs` before spawning — this path bypasses puppeteer entirely and was the last uncovered first-run trigger.
- **No registry sweeping.** The harness never mutates the user's Run key — an entry there may be intentional. Prevention at launch is the whole mechanism.
- `test/unit/e2e/launchPrefs.test.mjs` pins the const, the seeder (fresh write, idempotence, preserves foreign user.js content), and the detached-spawn wiring (the E2E legs are path-filtered CI, so nothing routinely executes these on PRs).
- Verified on the dev machine's real Firefox: a fresh-profile harness launch (the exact first-run trigger) left the HKCU Run key unchanged — zero new entries.

## Cleanup already done on the affected machine (outside this PR)

All 30 `Mozilla-Firefox-*` Run entries + matching `StartupApproved` values were deleted (HKCU only), plus the scheduled task left by a repo-local test install. Real browser registrations were never touched.

Fixes #191

### DIFF
--- a/test/e2e/updater/smoke-migration.mjs
+++ b/test/e2e/updater/smoke-migration.mjs
@@ -1,0 +1,311 @@
+#!/usr/bin/env node
+// test/e2e/updater/smoke-migration.mjs — ONE-OFF smoke test (untracked, not a
+// repo gate): drives the ADR 0026 dead-dev-channel fallback in a real Firefox.
+//
+// Installs NOTHING: it uses the machine's existing Firefox (discovered, or
+// FIREFOX_BINARY) with a throwaway profile in %TEMP%, and sets
+// toolkit.winRegisterApplicationRestart=false so the browser never registers
+// itself in the user's Windows startup (HKCU Run) on exit.
+//
+// Setup:
+//   - discovered local Firefox + disposable %TEMP% profile
+//   - profile seeded with the REAL dev-build-main-450468f utils (IS_DEV,
+//     STABLE_* baked) and fx-folder into the browser's GreD (see the GreD
+//     warning below — read before running on a personal machine)
+//   - override prefs: HASHES_URL → a dead file:// path (the deleted branch),
+//     STABLE_* + UI_BASE_URL → a local HTTP server serving the branch's own
+//     manifest and zips under stable names
+//   - utils forced stale so the check finds an update and opens the tab
+//
+// Asserts: fallback logged, migration persisted (activeChannel +
+// activeChannelBuild), tab opened, migration banner visible, the local stable
+// server served hashes.json + utils.zip, and the installed file was replaced.
+//
+// Usage: node test/e2e/updater/smoke-migration.mjs [--headless] [--firefox <path>]
+
+import fs from 'node:fs';
+import http from 'node:http';
+import os from 'node:os';
+import path from 'node:path';
+
+import {launchFirefox, attachProcessLogging, tempDir, rmDir} from '../shared/helpers.mjs';
+import {findGreDir, extractZip, discoverFirefoxBinary} from '../shared/browsers.mjs';
+import {closeBrowser, removeProfileCompatibilityIni} from '../shared/processHygiene.mjs';
+
+const BRANCH = 'dev-build-main-450468f';
+const RAW = `https://raw.githubusercontent.com/onemen/firefox-scripts/${BRANCH}`;
+const PORT = 8791;
+const HEADLESS = process.argv.includes('--headless');
+const FIREFOX_FLAG = process.argv[process.argv.indexOf('--firefox') + 1];
+const STALE_FILE = 'RDFDataSource.sys.mjs';
+const STALE_MARKER = '\n// fxs-smoke: forced stale\n';
+
+const work = fs.mkdtempSync(path.join(os.tmpdir(), 'fxs-smoke-'));
+const pkgDir = path.join(work, 'pkg');
+const serverHits = [];
+let server;
+
+const ok = [];
+const fail = [];
+function check(name, cond, detail = '') {
+  (cond ? ok : fail).push(name);
+  console.log(`  ${cond ? '✔' : '✗'} ${name}${detail ? ` — ${detail}` : ''}`);
+}
+
+async function fetchBranch(name, dest) {
+  const res = await fetch(`${RAW}/${name}`);
+  if (!res.ok) throw new Error(`${RAW}/${name} → HTTP ${res.status}`);
+  fs.writeFileSync(dest, Buffer.from(await res.arrayBuffer()));
+  console.log(`  downloaded ${name} (${dest})`);
+}
+
+/** Serve the branch manifest + zips under stable names; log every hit. */
+function startServer(dir) {
+  server = http.createServer((req, res) => {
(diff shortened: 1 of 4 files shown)


test/e2e/updater/smoke-migration.mjs
```diff
@@ -0,0 +1,315 @@
...
+import {launchFirefox, attachProcessLogging, tempDir, rmDir} from '../shared/helpers.mjs';
+import {findGreDir, extractZip, discoverFirefoxBinary} from '../shared/browsers.mjs';
+import {closeBrowser, removeProfileCompatibilityIni} from '../shared/processHygiene.mjs';
+
+const BRANCH = 'dev-build-main-450468f';
+const RAW = `https://raw.githubusercontent.com/onemen/firefox-scripts/${BRANCH}`;
+const PORT = 8791;
+const HEADLESS = process.argv.includes('--headless');
+const FIREFOX_FLAG_INDEX = process.argv.indexOf('--firefox');
+const FIREFOX_FLAG = FIREFOX_FLAG_INDEX === -1 ? '' : process.argv[FIREFOX_FLAG_INDEX + 1];
```
onemen: 🤖 AI review triage (Codebuff agent — result of the review:batch run)

**Valid.** `process.argv[process.argv.indexOf('--firefox') + 1]` with no `--firefox` flag evaluates `argv[0+1] = argv[0]` — the node executable path — so a missing flag silently "found" a bogus Firefox path instead of falling through to `FIREFOX_BINARY`/`discoverFirefoxBinary()`.

Fixed in fb1a587: index checked first; falsy when the flag is absent.

test/e2e/updater/smoke-migration.mjs
```diff
@@ -0,0 +1,315 @@
...
+  let polls = 0;
+  let tabFound = false;
+  let bannerLogged = false;
+  const watcher = Cc['@mozilla.org/timer;1'].createInstance(Ci.nsITimer);
+  watcher.initWithCallback({
+    notify() {
+      try {
+        if (++polls > 45) { watcher.cancel();
+          if (!bannerLogged) {
+            const msg = 'BANNER_VISIBLE unknown (timeout)\n';
```
onemen: 🤖 AI review triage (Codebuff agent — result of the review:batch run)

**Valid.** The `fos.write` length arguments were hardcoded (`34`, `.length + 2`) while the strings are dynamic — the BANNER timeout path could read past the buffer. Both write sites now derive the length from the actual string being written, matching the existing console-mirror pattern.

Fixed in fb1a587 (BANNER timeout + TAB_OPENED paths).

PR #194, opened 2026-09-13
feat(e2e): portable recipes for the NSIS fork installers (zen, floorp, waterfox)

onemen:
## What

Fork-portable expansion, phase 1 (#38, pulled into pre-1.0 by #4): the three fork browsers whose Windows installers are NSIS gain **portable install capability** — the same official installer, installed into `PORTABLE_BROWSER_DIR` via NSIS `/D=` instead of Program Files, exactly the mechanism the Firefox portable recipe already uses (no semantics change to any existing leg).

- `zen`, `floorp`, `waterfox` win recipes declare `portable: true` + `portableExe`.
- `PORTABLE_BROWSER_DIR` + a fork browser routes through the new `installForkPortable`: identical download chain (zen/floorp stable latest URLs; waterfox via the resolver's mirror → ci-downloads → cached-installer fallbacks, with vendor sha256 verification), `/S /D=<dir>` verbatim args (`nsisPortableArgs`), and a strict launcher-file check at `<dir>/<portableExe>`.
- `--url` output is unchanged → the fork-portable CI legs (phase 2) can reuse the existing URL-keyed download cache, and the extracted-dir cache from #193.

Vendor artifacts verified live (2026-09-13): Floorp stable latest ships `floorp-windows-x86_64.installer.exe` (NSIS); Zen stable latest ships `zen.installer.exe` (NSIS); Waterfox CDN serves `Waterfox Setup <v>.exe` (NSIS) — Linux/macOS already have tarball/dmg artifacts for a later phase.

## Deliberately out of scope

**No workflow change.** The CI legs (a fork-portable matrix job) land in phase 2 once these recipes are proven — including the open question of whether each NSIS installer actually honors `/D=` into a *fresh* directory (zen is an electron-builder NSIS; floorp/waterfox ship Mozilla-style NSIS). A failed assumption fails exactly one advisory leg then, not a recipe+workflow PR.

## Tests

Unit (3 new, suite 383→386, all green): recipe shape (`portable` + `portableExe` + args + url/resolver), `nsisPortableArgs` argv contract (`/S` then final `/D=`, spaces inside), and `--url` cache-key parity with the registered installs.

Part of #38 (pre-1.0). Part of #4.


.github/workflows/e2e.yml
```diff
@@ -447,13 +447,23 @@ jobs:
...
       - name: Resolve Firefox Release download URL (Linux/macOS)
         id: portable-url-unix
         if: ${{ runner.os != 'Windows' }}
         shell: bash
         run: |
           URL=$(node test/e2e/shared/downloads.mjs firefox --url)
-          KEY=$(node --input-type=module -e "import {createHash} from 'node:crypto'; console.log(createHash('sha256').update(process.argv[1]).digest('hex').slice(0, 16))" "$URL")
+          FINAL=$(curl -fsSI -o /dev/null -w '%{redirect_url}' "$URL")
+          test -n "$FINAL" || FINAL="$URL"
+          KEY=$(node --input-type=module -e "import {createHash} from 'node:crypto'; console.log(createHash('sha256').update(process.argv[1]).digest('hex').slice(0, 16))" "$FINAL")
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**major — portable cache key pins a stale Firefox version.** The portable legs keyed both caches (installer + extracted dir) on the hash of `?product=firefox-latest` — a redirect that never changes. `downloadTo`'s HEAD-size check self-heals the installer cache, but the extracted-dir cache has no such check: on a Firefox release, the old extracted dir is restored, the skip check finds its launcher file, and the leg silently keeps testing the old version.

Fixed in eeaaa10: the key now hashes the one-hop version-embedded CDN URL (read via the redirect's Location / `%{redirect_url}`), verified live for all three OS platforms.

test/e2e/shared/downloads.mjs
```diff
@@ -588,6 +701,72 @@ async function installDmg(url, appName) {
...
+  const exe = path.join(downloadDir(), `${browser}-portable-setup.exe`);
+  await downloadTo(url, exe);
+  if (sha256Url) {
+    console.log(`  verifying vendor sha256 for ${browser} (portable)`);
+    await verifySha256(exe, sha256Url);
+  }
+  // NSIS /D= consumes the rest of the command line (no quotes allowed), so
+  // spawn with verbatim args exactly like the Firefox portable recipe — and
+  // through the same AV file-lock retry (run 34741552124, floorp leg).
+  runNsisInstallerWithRetry(exe, nsisPortableArgs(dest), `${browser} portable installer`);
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**major — the portable NSIS spawns bypassed the AV file-lock retry.** `installForkPortable` (and the firefox-portable spawn above it) called `spawnSync` directly — the exact race that killed the floorp leg (run 34741552124): a portable leg downloads the installer and executes it immediately, which is the worst case for a Defender scan. A spawnSync failure also surfaces a different shape (libuv `error.code` = EBUSY, the os error 32 sharing violation) that the old predicate missed.

Fixed in eeaaa10: `runNsisInstallerWithRetry` wraps the spawnSync path (verbatim args and exit checks unchanged, `spawn`/`sleep` injectable for tests), and `isFileLockError` now matches the EBUSY shape — pinned by 4 new unit tests.

PR #195, opened 2026-09-13
installer: date-based self-update + gh-pages installer mirror (ADR 0019 amendment)

onemen:
## The bug

The installer's self-update compared the hardcoded `VERSION=1.0.0` against the `latest` release's `tag_name` with a plain strcmp. Under ADR 0019's release scheme, `latest` is a **permanently-named moving tag** — `strcmp("1.0.0", "latest")` can never be equal, so:

- today (with zero installer assets on `latest`) the check is dead code, and
- the moment installers ship as `latest` assets — which ADR 0019 requires — **every install would show the update banner forever**.

## The fix (ADR 0019 amendment, dated 2026-09-13 in the record)

**Detection is date-based, not version-based:**

- `config/installer.conf` gains `BUILD_DATE=2026-09-13` (replacing `VERSION`); the generated `CFG_BUILD_DATE` bakes it into every binary. ISO dates compare lexicographically = chronologically.
- The publish automation embeds a managed JSON block in each `installer-<date>` release body: `{"installerDate": "...", "download": {"<asset>": "<latest-tag URL>"}}` (rendered in a ```json fence, invisible to humans).
- `check_self_update()` parses the block (escaped-body aware, substring-safe: `installer_linux` never matches `installer_linux_aarch64`; assets' `browser_download_url` decoys are ignored by construction — the URL search starts after the date), compares `BUILD_DATE` vs `installerDate`, and returns the managed URL.
- The tab ingests the **newest `installer-<date>` body via `/releases?per_page=10`** (a scripts-only republish of `latest` can never mask a newer installer) with the `latest` release as fallback; **local/dev builds never check** (`selfUpdateDisabled` on `/api/build-info`).
- Banner shows both dates ("A new installer build is available: **2026-09-13** (this installer: 2026-09-01)") plus the instruction line from the mock annotation: *"Download the new installer, then close this window and run the downloaded file — it replaces the old one."* A newer build without a managed URL for the platform falls back to the releases page.
- `VERSION=1.0.0` removed; Windows VERSIONINFO carries the build date.

**gh-pages installer mirror** (user-reported gap: `gh-pages` had no installers): prod publishes now mirror `installer_*` to the Pages branch — the release asset stays the user-facing download; the Pages copy gives the banner's fetch a CORS-enabled host (release-asset CDNs send none). Helpers remain Pages-only, never release assets.

## Tests

- `installer/test/test_self_update.mjs` rewritten for the date contract (11 cases: same/older/newer dates, year boundary, managed-URL fallback, releases-listing decoy assets, top-level block, no block, malformed date) — runs in `pnpm test:hash` against a real built binary.
- Installer E2E **Test 6** now drives the endpoint verdict paths: newer-date → update + URL, same-date → up to date, no-block → silent, plus the date-not-version contract assertion.
- 6 new unit tests for `renderSelfUpdateBlock` / `parseSelfUpdateBlock` / `mergeSelfUpdateBlock` (same-day merge semantics, foreign-date drop).

## Docs

- ADR 0019 amended in place (dated amendment section) per maintainer decision — no new ADR.
- `docs/ci-inventory.md` unaffected (no workflow trigger/filter change; comment-only edit in ci.yml).


onemen (commented):

*Standalone line-anchored threads were rejected by the GraphQL API (permission error on `AddPullRequestReviewThread`), so this review comment carries the findings per the ADR 0020 fallback. All four accepted findings are already fixed on head `5162eda`.*

**1 · major — `installer/web/script.js` (ingest race) — FIXED in `5162eda`**
The legacy self-update POST in `ingestRemoteData` raced the new `ingestSelfUpdateSources` and could overwrite the managed installer payload with a block-less latest-release body — the banner would silently never show. Removed; self-update ingestion now has exactly one owner (releases-list first, `latest` fallback).

**2 · major — `test/e2e/installer/installer-e2e.mjs` (fixture date) — FIXED in `5162eda`**
The hard-coded fixture date would eventually equal (or pass) the binary's baked `BUILD_DATE` and flip the expected verdicts. Test 6 now probes the binary's `buildDate` first and derives `dayBefore`/equal/`dayAfter` fixtures from it (adding the ahead-of-binary case).

**3 · minor — `tools/publish/componentReleases.mjs` (bare-block extraction) — FIXED in `5162eda`**
The bare-block fallback regex (`[^{}]*`) truncated the nested `download` map, so a same-day merge would lose prior platforms' URLs. Replaced with backwards-walk-to-opening-brace + balanced-brace extraction; round-trip test added.

**4 · minor — `installer/src/self_update.c` (stale ADR ref) — FIXED in `5162eda`**
Stale "ADR 0027" references corrected to the ADR 0019 amendment wording (maintainer chose amendment-only) in `self_update.c` and `self_update.h`.

**5 · major — managed-block URL host — SKIPPED (wrong)**
The finding asked for Pages-host URLs in the managed block. The managed URL intentionally targets the `latest` **release** download (the user-facing artifact; the banner downloads via an anchor click, which needs no CORS). The gh-pages mirror exists for fetch-based flows, not as the banner's target. A comment now documents this.

onemen:

Four accepted findings from the earlier batch were fixed and pushed (5162eda), and CI on that head exposed two more Test 6 bugs — both fixed and pushed (6b59dd8), each verified locally against a fresh Windows binary (42/42 HTTP-layer checks):

1. Inverted fixture semantics: Test 6 expected an update when the published installerDate was OLDER than the binary; the C contract (self_update.c) is the reverse — update only when the published date is newer. Fixtures now assert: newer → update + managed URL, equal → up to date, older → silent.
2. Date arithmetic bug: `new Date(x) + 24*3600*1000` triggers Date ToPrimitive (string concat) and silently yields the SAME day — this is why CI's latestDate equaled the build date. Derived from Date.parse() timestamps now.
3. (Also) the managed download map key now comes from /api/build-info's assetName instead of a hard-coded -dev suffix, so the leg validates against prod and dev binaries alike; build-info gained assetName/buildDate fields.

Remaining (dismissed/wrong): Pages-host URL in the managed block is intentional — the banner uses an anchor click, no CORS needed.

PR #196, opened 2026-09-13
e2e: fail fast instead of hanging — watchdog, orphan sweep, stale-binary guard

onemen:
## Problem

Installer E2E runs stall **indefinitely on local Windows machines** while passing in CI (observed repeatedly on the maintainer's machine during #195 validation). Diagnosis found three compounding failure modes:

1. **No global watchdog.** Every leg bounds its own awaits, but a wedged child (a GUI MessageBox on a dev machine, an OS-level stall) hangs the runner until something external kills it — locally that means a 10-minute harness timeout with no diagnostics, in CI a cancelled job.
2. **Orphaned installers poison later runs.** When a run is killed (timeout or Ctrl-C), its spawned installers are *not* children Node reaps — they survive (`installer_win.exe` PIDs observed holding `127.0.0.1:8777` for hours). Every subsequent run's HTTP layer and TS-3/TS-4 legs then probe the **dead run's server**, producing phantom passes or hangs. The existing `killStrayProcesses()` pre-run sweep helps the *next* run, but nothing swept on the way *out* of a throwing/timeout run.
3. **Stale snapshot binaries crash the runner.** After #195, Test 6 parses `/api/self-update`'s `buildDate` and `/api/build-info`'s `assetName`. A snapshot built before that contract reports `null` for both; `Date.parse("nullT12:00:00Z")` → NaN → `toISOString()` throws `RangeError` → the whole runner dies mid-suite with no explanation. Reproduced with the Sep 4 `prod-main-01dccd8` snapshot.

## Fix

All in `test/e2e/installer/installer-e2e.mjs`:

- **20-minute global watchdog** — sweeps stray installer children then exits non-zero, so a wedged run kills its own orphans instead of leaking them.
- **Orphan sweep on every exit path** — success, mid-run throw (`run().catch`), and watchdog expiry. The pre-run sweep remains as belt-and-braces for external kills.
- **Stale-binary guard in Test 6** — when `assetName`/`buildDate` are missing or malformed, fail one explanatory check ("binary predates the assetName/buildDate contract — rebuild the snapshot") and skip the date fixtures instead of throwing.

## Validation

- Stale-binary run (`prod-main-01dccd8` + current test): degrades to `31/34` with the rebuild hint and a clean exit — previously a `RangeError` crash.
- Full run with the current binary on Windows: **57/57 PASS** including TS-3 and TS-4, and the new post-run sweep logs "no stray processes" — zero leaked installers.
- `pnpm lint` / `pnpm format` clean; `pnpm test` 404 tests, 0 failures.

Test-only change (no core/** files), so no new core test obligation per the CI gate.


onemen:

1 finding from the batch, assessed **valid** and fixed in c358c41: the stale-binary guard accepted an empty/partial `assetName` (the earlier `assetName.startsWith(plainBase)` check failed but execution continued with a broken download-map key) and regex-valid impossible dates (`2026-99-99` → `Date.parse` → NaN → the exact RangeError the guard exists to prevent). The guard now requires the `plainBase` prefix and a parse that round-trips to the same UTC day. Re-validated: stale-binary run 42/42 with the guard skipping the date fixtures cleanly.

Issue #197, opened 2026-09-14
Updater E2E speedup: merge stale-variant scenarios into one browser session (profile attached)


onemen:
### What

The updater E2E runs 8 scenarios serially, each paying a full Firefox launch (~5–8 s on `windows-latest`). Three of them (`utils-stale`, `config-stale`, `both-stale`) differ only in which daily-gate pref is stale; they could run in **one** browser session with pref mutations between checks.

### Measured profile — `portable Firefox E2E · windows-latest`, run 34760339728 (373 s total, 2026-09-13)

| Phase | Duration | Share |
|---|---|---|
| Setup: checkout + caches + pnpm + native installer build | 79 s | 21% |
| Portable Firefox install (cache hit) | 22 s | 6% |
| installer E2E (`--ui --headless`, 76/76) | 77 s | 21% |
| **updater E2E (31/31)** | **224 s** | **61%** |

Updater scenarios (windows-latest):

| Scenario | Duration |
|---|---|
| utils-stale | 25 s |
| config-stale | 25 s |
| both-stale | 22 s |
| up-to-date | 8 s |
| skipped | 11 s |
| install-applies | 43 s |
| manual-install-upgrade (2 browser sessions) | 48 s |
| manual-install-no-ui (2 sessions) | 39 s |

Single Firefox launch: 4.6–7.6 s (harness `[timing]` lines). The leg's cost is dominated by ~11 serial browser launches — the deliberate price of per-scenario isolation.

### Proposed optimization

Merge the three stale-variant scenarios into one browser session (pref mutation between checks): saves ~35–40 s (~10% of the job). Requires reworking the harness's daily-gate pref handling (`lastScriptsCheckDate` / `lastUpdateTabShown`) to mutate prefs in-session instead of reseeding.

**Estimated win is small.** Do this only alongside the bigger E2E speed lever: **caching the fork-browser downloads** (zen/waterfox/floorp/librewolf re-download their ~100–200 MB browser every run — minutes, not seconds).

### Non-goals

- Do not merge `install-applies` / `manual-install-*` — those need real install events and two-session isolation.
- Do not tighten job timeouts below current values (20 min on these legs is ~2.5× worst-case cold-cache, see docs/ci-inventory.md → Job timeouts).

onemen:
Documenting a live flake observed after this issue's body was written, and folding it into the implementation plan.

[...]

PR #198, opened 2026-09-14
ci: explicit job timeouts + tunable installer E2E watchdog

onemen:
Closes the timeout half of #197's context: every `ci.yml`/`e2e.yml` job gets an explicit `timeout-minutes`, and the installer E2E's global watchdog is tightened and made tunable.

## Why

GitHub's default job timeout is **360 minutes** — a wedged job (GUI dialog on a dev machine, OS stall) hangs for six hours before cancellation. The installer E2E got a 20-min in-script watchdog in #196, but nothing bounded the other ~20 jobs, and 20 min was generously sized.

## What

**Workflow timeouts** (`~2.5–4x` the slowest green run on the slow legs; generous minutes-scale floors on cheap jobs — measured 2026-09-13, see the new "Job timeouts" section in `docs/ci-inventory.md`):

| Job(s) | Timeout | Slowest green observed |
|---|---|---|
| `detect changed paths` (both) | 5 min | 7 s |
| lint + format, gates, post-gate jobs | 10 min | 28 s |
| publish gate `<os>` | 20 min | 129 s |
| snapshot / installer E2E / helper | 15 min | 143 s |
| updater / portable / snap / browser-matrix / waterfox | 20 min | 373 s |

**Installer E2E watchdog**: 20 → **10 min** (4× its slowest green run) with an `E2E_WATCHDOG_MIN` env override for slow machines. It must stay below the job timeout so the script's orphan sweep + `[watchdog]` log always wins the race — CI shows the graceful cleanup, never a bare VM kill.

## Validation

- Env-parse sanity checked (unset/empty/garbage → 10 min; `0` → 10 min; `3` → 3 min).
- `pnpm lint` 0 issues, `pnpm format` clean, `pnpm test` 404/0.
- No behavior change on green runs — the watchdog never fires healthy; timeouts only cap wedges.

Part of #197.


onemen:
🤖 AI review triage (Codebuff agent — CodeRabbit review:batch run on PR head)

Two batches ran on this PR (head `2557be9` and head `58cd36c` after fixes). Findings, each assessed per ADR 0020:

**Batch 1 (2 findings):**

1. *"~3× timeout ratio is not supported by the measured values (5 min vs 7 s is 43×)"* — **accepted (minor, docs)**. Fixed in `00cf4f5`: the claim now states the ratio holds for the slow E2E legs (the margins that matter, ~2.5–4×) while cheap jobs get a generous minutes-scale floor by design. PR body corrected too.

2. *(scrolled out of the first capture; re-run confirmed only one remaining finding, so this was the same docs point or subsumed by it — nothing else was ever reported against the workflow files).*

**Batch 2 (1 finding):**

2. *"E2E_WATCHDOG_MIN=0 selects the 10-minute default instead of the documented 1-minute floor"* — **accepted (minor)**. `parseInt('0')` is finite, so the old `|| 10` fell through to the default before `Math.max` could clamp. Fixed in `58cd36c` with the review's proposed parse-then-clamp form; semantics re-verified (8 cases: unset/empty/garbage → 10, `0`/`-4` → 1, `3` → 3).

No unresolved findings. Both fixes verified locally (lint 0, format clean, 404 unit tests 0 fail) and by the full CI+E2E matrix on the final head.

PR #199, opened 2026-09-14
e2e: fork-portable legs — NSIS installers into a custom directory (#38)

onemen:
Part of #38 — the pre-1.0 fork portable E2E legs. C1 of the SignPath-wait plan (task-map review).

## What

New **advisory** `fork-portable` E2E job (windows-latest, matrix: zen, floorp, waterfox): the official NSIS installer runs `/S /D=<runner-temp>/fork-portable` — the exact registered installer, but into a custom directory — and the updater E2E then runs against that **unregistered** binary. This covers the portable layout a large share of fork users actually run, which no existing leg exercises (browser-matrix and updater-waterfox install into Program Files).

## Why these three

`#194` already added `portable: true` + `portableExe` recipes and the `/D=` mechanism for zen/floorp/waterfox and verified their installers honor `/D=` into a fresh directory. LibreWolf's setup does not declare that capability — its leg stays registered-only, and a single-browser `browser=librewolf` escape **skips this job** (job `if:` and the gate's applicability entry carry the same predicate, so the gate contract stays exact).

## Implementation

- `downloads.mjs` `installForkPortable` gains the same cache skip as `installPortableFirefox` (`statSync().isFile()` at `dest/<portableExe>`), and the job caches the extracted dir under the same URL-derived key family (`<key>-x`) as the portable Firefox leg — invalidates with the browser version, so a stale extracted dir can never be validated in place of the new release. A warm run skips the ~100 MB download **and** the silent install.
- Gate wiring: `fork-portable` added to e2e-gate `needs` (advisory), `applicability`, `results`, and `skip-guard`; `tools/check-gate-coverage.mjs` contract extended — `pnpm check:gates` holds at 13 jobs.
- `BROWSER_PIN_VERSION` dispatch input passes through, so the ADR 0025 waterfox break-glass pin exercises the portable path too.
- `docs/ci-inventory.md` updated (job row + gate semantics).

## Validation

- `pnpm test` 406/0 (2 new unit tests: matrix browsers must keep portable capability; cache-skip predicate rejects a directory at the launcher path).
- `pnpm lint` 0 issues · `pnpm format` clean · `pnpm check:gates` ✓ · `pnpm check:decisions` ✓.
- The legs themselves run for the first time in this PR's E2E run.

Fixes part of #38.


onemen:
🤖 AI review triage (Codebuff agent — CodeRabbit review:batch run on PR head)

One batch ran on head `fd782e3`. One finding:

1. *"Extracted-dir cache key is not release-specific: zen/floorp use stable latest/download URLs, so the URL hash never invalidates — the leg could validate a stale browser."* — **accepted (major)**. Confirmed empirically first: the registered zen leg on main re-downloads every run (`downloadTo`'s HEAD-size self-heal), so the bug class was latent, not active — but the extracted-dir cache this PR adds would have made it real. Fixed in the follow-up commit with two layers: (a) the workflow hashes the version-bearing redirect target after one HEAD hop (the portable Firefox leg's exact pattern), and (b) `installForkPortable` re-proves the cached extracted dir against the cached installer's size vs the remote content-length, falling through to a fresh install on mismatch or when the installer cache entry is gone.

No unresolved findings. Gates re-run after the fix: 406/0 tests, lint 0, format clean, check:gates ✓.

PR #200, opened 2026-09-14
docs(readme): finding your version and reverting (date-based packages)

onemen:
Part of #4's "Version visibility" item: per-package dates verified in the installer + updater UI, plus a README "find your version / revert" section.

## What

A single README section, **"Finding your version and reverting"** (after the packages section, before the license):

- Explains the date-based package identity (ADR 0019): each package carries the date its content last changed (`YYYY-MM-DD`, UTC) — there are no version numbers.
- **Where to see it**: the updater tab and the installer both show the published package dates next to the manual-download links (`fx-download-date` / `utils-download-date`, rendered by `setDownloadDate` in `updater.js` / `installer/web/script.js`), with Up to Date / Update Available status per browser.
- **How to revert**: `latest` always holds the newest build; the frozen `scripts-<date>` / `installer-<date>` component releases (#176, ADR 0019) hold each past publish. Download the zip from the dated release and install it by hand (manual download link, or unpack into `chrome/utils/`), then use the per-update "Don't show again for this update" skip checkbox (`extensions.firefox-scripts.skippedHash.*`) to stay on the older build until the next published change.

## Verification

- Every claim checked against the code: date rendering in both UIs, the skip pref + checkbox, and the component-release naming from the publish automation.
- Note: `latest` currently has no assets (the next `pnpm release` materializes the mirror — #195 machinery is merged, the run is pending), so the revert flow becomes fully usable on the next publish. The section describes the shipped behavior, not the current empty state.
- Lint + format clean; docs-only change (no CI path filters triggered beyond docs).

Closes part of #4.

PR #201, opened 2026-09-14
e2e: retry the AV file-lock race on every silent-install path (#4)

onemen:
Fixes the E2E-resilience checkbox in #4: retry the silent-install step against the Windows AV file-lock race.

## The bug

Windows Defender intermittently holds a file lock on the freshly downloaded `setup.exe` for the first seconds after the download completes. Three legs died to it before any test code ran:

- floorp registered leg, 2026-09-13 (run 34760231467 — the reason this checkbox exists)
- registered floorp + nightly legs, 2026-09-14 (run 34808851430 job 103866152255: "The process cannot access the file because it is being used by another process")
- same-day PR #199's floorp run (34814023362)

The portable recipes already ride this out (`runNsisInstallerWithRetry` → `runInstallerWithRetry`), but the **registered** paths still called `execSync` directly with no retry.

## The fix

All five silent-install call sites now go through one `runSilentInstaller` wrapper backed by `runInstallerWithRetry` (4 attempts, 4 s apart, lock-signature gated by `isFileLockError`):

- `installInstaller` main path (firefox/firefox-dev/nightly registered NSIS)
- `installInstaller`'s cached-installer fallback
- resolver path (librewolf/waterfox) main install + both cached fallbacks

A non-lock failure (real NSIS error) still fails immediately — no retry noise.

## Test

New unit test pins the **exact error shape the failing CI runs produced**: `execSync` with `stdio: 'inherit'` surfaces the scanner's text only in `err.message` (`err.stderr` is null — verified locally and consistent with the CI log, where the signature is printed to the log by the inherited stderr and the thrown error carries it in the message). The retry consumes that message-only signature and succeeds.

`pnpm test` 407/0 · lint 0 · format clean.

Closes the "E2E resilience" item in #4.


onemen:
🤖 AI review triage (Codebuff agent — CodeRabbit review:batch run over #200 + #201 in one batch)

**No findings.** Both PRs reviewed clean on their heads (`43f1e70` for #200, `5f81859` for #201): README.md, downloads.mjs, and the unit test file produced zero review comments.

Nothing to triage, nothing to fix; per ADR 0020 the empty result is recorded here with the 🤖 provenance marker.

PR #206, opened 2026-09-15
fix: land the audit's three P0 findings (timeouts, route contract, docs drift)

onemen:
Implements the three P0 findings from the 2026-09-15 comprehensive audit (the 2026-09-15 working-tree audit (local draft, not committed), local draft, quoted for provenance).

Part of #4 (Phase 5 housekeeping). Follow-up queue for the P1/P2 findings lives in the audit doc.

## The three P0 findings, as audited

> **P0 (Critical) | CI/CD** — `pages.yml` prod publish jobs have no `timeout-minutes` while `pages-publish` is non-cancellable. *Add `timeout-minutes` to all 5 jobs (20-30 min, matching `build-and-upload.yml`); same for `cache-cleanup.yml` and the watchdog `pr-check` jobs.*
>
> **P0 (Critical) | Security** — `/api` token-gate coverage is a hand-maintained list (`smoke-security.mjs:38-49`). *Add a unit test diffing the routes registered in `installer/src/main.c` against the smoke test's gated/open/shutdown sets; fail on any unclassified route.*
>
> **P0 (Critical) | Docs** — `docs/future-work.md` §6 and §2.2/§3 claim shipped features are unimplemented (test flags, `env.json`, hygiene, helper sidecar). *Rewrite §6 to open items only; strike the completed checkboxes with the shipping commit/PR references.*

## What each commit does

**1 · `ci: explicit timeout-minutes on every publish/watchdog job`**
- `pages.yml`: timeouts on all 5 jobs (10/10/20/20/20) + a rationale comment on the non-cancellable concurrency group. A wedged job previously held the publish slot for GitHub's 360-minute default.
- The audit's same-item tail, verified still open and now closed: `cache-cleanup.yml` (15), `skills-watchdog.yml` check/pr-check (15/10), `url-watchdog.yml` `pr-check` (15 — half of `check`'s 30).
- `docs/ci-inventory.md` timeout section extended to all four workflows (AGENTS.md requires the inventory stay in sync).

**2 · `docs: rewrite future-work shipped-features sections to open items only`**
- The "verified not implemented as of 2026-09-05" banner and its §2.2/§3/§6 rows corrected: `--port 0`/`--server-only`/`--env-file`, `env.json`, E2E process hygiene and the helper `.sha256` sidecar all shipped (D1/D2); the `msys2 update:` row now states the per-workflow truth (D3). Each struck item carries its shipping reference.

**3 · `test: turn the /api token-gate list into a checked contract`**
[...]

PR #207, opened 2026-09-15
test: stale-snapshot guard (T1) + gitignoreUtils coverage (audit P1 batch 1)

onemen:
Lands the first two P1 items from the 2026-09-15 comprehensive audit (the 2026-09-15 working-tree audit (local draft, not committed), local draft, quoted for provenance). Test-only change; no production code touched.

Part of the audit follow-up queue (after #206's P0 batch). Per the operator's plan, review for this and subsequent audit PRs happens as a batch when all are ready.

## P1 · T1 — `test_self_update.mjs` runs an arbitrary binary

> **P1 (High) | Testing** — `test_self_update.mjs` runs whatever binary sits in the newest snapshot dir, with no freshness check (T1). *Compare `hashes.json.installer.hash` against a fresh `computeFileSetHash`; rebuild or fail with an explicit "stale snapshot — run `upload:local`" message.*

**Implemented exactly that:** before running any case, the suite recomputes the installer's source-tree hash the way `upload.mjs buildBinaries` does (installer/src minus `helper/**` and the generated headers, installer/web, `config/installer.conf`) and diffs it against the snapshot's recorded `installer.hash`. On drift:

```
STALE SNAPSHOT: dist\dev-main-83664f0
  its installer was built from sources with hash a5c408d5…,
  but the current sources hash to           747d2ce1….
  (snapshot installer date: 2026-09-07)
Refusing to run the self-update suite against binary provenance this
suite cannot trust — failures would misreport as regressions on main.
Regenerate: pnpm upload:local -- --mode=prod   (or --mode=dev for a local snapshot).
```

**Both directions validated live:**
- Stale path: pointed the suite at the workstation's real stale snapshots (built 2026-09-07/09-10, pre-dating the 2026-09-13 self-update change) → exit 1 with the message above — the exact scenario that produced the audit's 17 misleading failures.
- Fresh path: `pnpm upload:local -- --mode=dev` in this branch, then `pnpm test:hash` → hash parity PASS (utils + fx-folder, JS == C) + `✓ all self-update checks passed`, with a one-line provenance note (`Snapshot provenance verified: installer built from current sources (747d2ce1da02, date 2026-09-13)`).

Implementation note: the publish modules call `requireMode()` at import time, so the test adopts the repo's existing `process.argv.push('--mode=prod')` arrangement (same as `zipContents.test.mjs` / `createZip.test.mjs`) — import-time only, nothing publishes.

## P1/P2 boundary · `gitignoreUtils.mjs` test coverage

> **P2 (Medium) | Testing** — `gitignoreUtils.mjs` (zip + hash input selection) and `syncGeneratedFiles.mjs` have no tests. *Add unit tests: pattern precedence…*

This module decides what ships in every zip and what the publish hash covers — a silent bug changes what users install. 14 tests pin the parse/merge semantics, `shouldIgnore` matching (basename at any depth, path patterns, trailing-slash dirs, `**`, Windows separators), and `getAllFiles` behavior (pruning, `includeSubfolders`, `.git` auto-excluded while `.github` is walked).

Two non-gitignore semantics are pinned deliberately so nobody "fixes" a caller into git semantics:
- **First match wins** — the opposite of real gitignore's last-match-wins.
- **A trailing-slash directory match decides immediately** — a later `!dir/file` negation cannot rescue a file inside an ignored tree (the branch returns on the directory, and the walk prunes without descending). Verified against the implementation during test-writing; an *earlier* negation does win.

## Validation

- `pnpm test` — 423 tests, 420 pass, 3 skip (the +17 are the new gitignoreUtils tests; suite was 406)
- `pnpm lint` · `pnpm format` — green
- `pnpm test:hash` — full pass on a fresh snapshot (see above)

🤖 Generated with Codebuff


test/unit/publish/gitignoreUtils.test.mjs
```diff
@@ -0,0 +1,207 @@
...
+  // The trailing-slash branch RETURNS on the directory itself, before later
+  // (negation) patterns are consulted — and getAllFiles prunes the dir without
+  // descending. Unlike real gitignore (!logs/README would re-include), a file
+  // under a dir-ignored tree is unreachable by negation. Callers must express
+  // such exceptions as earlier sibling patterns, not later negations.
+  // Fixtures use the parser-normalized shape: parseGitignore strips the
+  // leading '!' into isNegation, so a stored pattern never contains '!' —
+  // passing '!logs/README' here would exercise minimatch's own inverted-match
+  // semantics instead of shouldIgnore's (batch review, 2026-09-15).
+  const patterns = P([['logs/'], ['logs/README', true]]);
```
onemen, line 108: 🤖 **AI review — agent-posted (ADR 0020)** — triage of the 2026-09-15 batch review (CodeRabbit CLI, one pass over PRs #206–#209) — accepted (right).

**Negation fixtures must not carry the leading `!`.** `parseGitignore` strips it into `isNegation` (`line.slice(1)`), so `{pattern: "!logs/README", isNegation: true}` never occurs from real input — and the "earlier negation wins" assertion was passing through minimatch's own inverted-match semantics, not `shouldIgnore`': verified live, `minimatch("README", "!logs/README") === true` (minimatch treats `!x` as NOT-match), so the basename check accidentally rescued the path.

**Fixed in 6f300a5** — fixtures use the parser-normalized shape (`{pattern: "logs/README", isNegation: true}`); all 15 assertions hold through the intended first-match path. Resolving on that commit.

PR #208, opened 2026-09-15
Pin the manifest hash order to an explicit case-insensitive comparator (audit P1)

onemen:
Unify the manifest hash comparator across C and JS, with adversarial parity tests

Part of #4 (P1 batch from the 2026-09-15 audit). Test-only + comparator-swap; no behavior change for the published ASCII file sets (verified — the probe and both real packages hash identically before/after).

## Finding (audit §3.3, P1)

The hash contract (ADR 0002) requires the C installer, the publish tooling and the in-browser updater to derive the **identical sorted file order**. They didn't specify it:

- C (`compute_directory_sha256`, `detect_browser.c:362`): `strcasecmp()` — C-locale byte order
- JS ×4 (`hashUtils.mjs:103,146`, `scriptsUpdater.sys.mjs:672,752`): `localeCompare()` — **locale-aware**; in Firefox that's the application locale, and locale order differs from byte order for `_`/`-`, digits, case adjacency

Latent today (audit probed 13 locales × real file lists — no divergence on current ASCII names), but any future non-ASCII or punctuation-heavy path would make publisher, installer and updater disagree about "up to date" — per-machine.

## Change

One explicit comparator everywhere, mirroring `strcasecmp`:

| Site | Before | After |
|---|---|---|
| `installer/src/detect_browser.c` | inline `strcasecmp` insertion sort | `cmp_path_ci()` — ASCII case-fold, unsigned byte compare |
| `tools/publish/hashUtils.mjs` (×2 sites) | `localeCompare` | exported `compareCaseInsensitive()` (strings or `{relative}`/`{rel}` entries) |
| `core/chrome/utils/updater/scriptsUpdater.sys.mjs` (×2 sites) | `localeCompare` | local `compareHashOrder()` twin (sys module stays self-contained) |

`localeCompare` remains only where hash order is irrelevant (skills-watchdog inventory display).

## Tests (`test_hash.mjs` adversarial probe)

1. **Order stability** — 11 pairs chosen where naive orders disagree: case adjacency (`A.txt` vs `a!.txt`), digit-vs-letter (`config2.js`/`config10.js`/`configX.js`), punctuation (`file_1`/`file-1`, `x.js`/`x~.js`), extension shells (`utils.d.ts`/`utils.js`), dir/file case collisions (`ARM/` vs `a.txt`, `b/` vs `B.txt`). Each pair must sort identically in both input orders.
2. **Byte parity** — synthetic 12-file adversarial tree (case-colliding basenames across dirs, digits, punctuation), manifest written **shuffled**: the C `--test-hash` output must equal JS `computeFileSetHash` byte-for-byte.

## Validation

- `pnpm test:hash` green on a fresh build: real utils + fx-folder parity, 11/11 pairs, C == JS on the adversarial tree
- `pnpm test` (406 pass / 3 skip), `pnpm lint`, `pnpm format` green
- Note: hashUtils' sort semantics (entry objects, not strings) are now pinned by its existing unit tests — my first comparator draft passed strings only and they caught it immediately.

Timeout-tail housekeeping and the config batch are separate PRs (batch review planned once all P1 PRs are up).


tools/publish/hashUtils.mjs
```diff
@@ -17,6 +17,43 @@ import {warn} from './log.mjs';
...
+ * Accepts plain strings, or `{relative}` / `{rel}` entry objects as used by
+ * computeDirectoryHash / computeFileSetHash.
+ *
+ * @param {string | {relative?: string; rel?: string}} a
+ * @param {string | {relative?: string; rel?: string}} b
+ * @returns {number} negative / 0 / positive, for Array.prototype.sort
+ */
+const FOLD_OFFSET = 'a'.charCodeAt(0) - 'A'.charCodeAt(0);
+function foldC(bytes) {
+  const out = new Uint8Array(bytes.length);
```
onemen, line 41: 🤖 **AI review — agent-posted (ADR 0020)** — triage of the 2026-09-15 batch review (CodeRabbit CLI, one pass over PRs #206–#209) — accepted (right).

**The comparator contract should be byte-exact, not approximately aligned.** Two defects in one: (1) the JS side folded with Unicode toLowerCase() and ordered by UTF-16 code units while C folds ASCII bytes per byte — a divergence merely documented rather than eliminated, and Unicode folding equates pairs the C side keeps distinct (e.g. ß vs ss); (2) case-fold-equal distinct paths (A.txt vs a.txt) compared equal on both sides, making the sort order input-dependent — poison for a hash contract.

**Fixed in d914c48** — both twins now fold ASCII letters over the UTF-8 bytes (primary key) and tie-break folded-equal distinct paths on the raw bytes (C: strcmp; JS: Buffer.compare / raw loop). Verifier evidence: 361-pair cross-check of the two JS twins, 0 sign mismatches; total order on tie-break pairs; C == JS byte parity on the extended 13-file adversarial tree against the real binary. Resolving on that commit.

PR #210, opened 2026-09-15
fix: per-connection read deadlines on the installer HTTP server (audit P1 reliability)

onemen:
Implements the last untouched P1 finding from the 2026-09-15 comprehensive audit (the 2026-09-15 working-tree audit (local draft, not committed) §3.3 / §7 — Reliability). Part of #4 (Phase 5 housekeeping — audit follow-ups).

## The finding

> `http_server.c` — single-threaded blocking `accept()` loop; no per-connection read timeout. A local client that connects and dribbles bytes wedges the installer UI indefinitely. Loopback-only lowers severity; a socket timeout is cheap.

Because the serve loop is single-threaded, one stalled connection blocks **every** request behind it — including the UI polling `/api/status` and the browser's zip uploads. ADR 0010 covered who may drive the API (token gate) but not whether it stays reachable.

## The fix — two independent deadlines

An idle timeout alone does not close the finding: a client that dribbles one byte every few seconds resets `SO_RCVTIMEO` forever. Both bounds are applied:

- **Idle (10 s):** `SO_RCVTIMEO` on every accepted socket (plus the listen socket on Winsock, where the accepted handle does not reliably inherit it). Aborts connections that send nothing at all.
- **Total (30 s):** a monotonic-clock deadline over the whole request-read phase, checked after every successful recv. Closes the dribble case.

On expiry the request is answered **`408 Request Timeout`** and closed. Semantics of the read loop exit:

| `recv` result | Meaning | Response |
| --- | --- | --- |
| `n > 0`, total bound expired | dribble / very slow body | 408 |
| `n < 0` (error or idle expiry) | stalled connection | 408 (best-effort send) |
| `n == 0` | client closed cleanly mid-request | silent close (unchanged) |

`408` is deliberately a *response*, not just a close, so a monitoring client can distinguish "server shed you" from a network failure. The send is best-effort and no-ops if the peer is already gone.

## Test-only override

`http_server_set_timeouts()` (new, declared in `http_server.h`) is honored **only under `--smoke-test`** (`main.c` reads `FXS_HTTP_RECV_TIMEOUT_MS` / `FXS_HTTP_REQUEST_TOTAL_TIMEOUT_MS` there). Production always runs the defaults; the values are not parseable from any request surface.

## New smoke check (`smoke-security.mjs`)

- Opens a raw `net.connect` socket, sends nothing, and asserts the server answers `HTTP/1.0 408 Request Timeout` within 10 s (idle deadline shortened to 1.5 s via the override).
- Asserts the server **still serves** afterwards — a full valid-token `/api/claim` round-trip (the whole point: shed the stalled connection, keep the loop alive).
- Harness now spawns the installer with `FXS_HTTP_RECV_TIMEOUT_MS=1500`.

## ADR 0010

Added an **Availability note** beside the entropy note: records the single-threaded availability envelope, the two deadlines, the 408 semantics, and the smoke-only override mechanism.

## Validation

- Security smoke against the fresh dev-snapshot binary: **86/86 checks, exit 0** (84 existing + 2 new).
- `pnpm test:hash`: full pass (JS == C hash parity + self-update suite) against the new build.
- `pnpm test`: 428 pass / 3 skip; `pnpm lint` (incl. `gcc -fanalyzer`, clean) + `pnpm format` green; `pnpm check:decisions` OK.
- The intermediate defect where the idle expiry closed **without** the 408 was caught by the new check itself (first run: 85/86) and fixed in `bc06ef2`.


docs/decisions/0010-session-token-no-cors.md
```diff
@@ -27,4 +27,6 @@ ephemeral ports plus Origin checks) is adopted.
 **Entropy note (2026-09-15):** the token is a true 128-bit value — 16 CSPRNG bytes, both nibbles
 taken (`hex[raw[i] >> 4]`, `hex[raw[i] & 0xF]`). Earlier builds consumed only the low nibble
 (`raw[i] % 16`), yielding 64 bits from the same 16 random bytes; the smoke-security and E2E
-harnesses pin the 32-hex length.
+harnesses pin the 32-hex length.**Availability note (2026-09-15):** the single-threaded serve loop's
+availability envelope — per-connection read deadlines, 408 semantics, and the smoke-test override —
+is its own decision: see [ADR 0028](./0028-installer-serve-loop-availability.md).
```
onemen, line 32: 🤖 AI review triage (Codebuff agent — result of the batch `review:batch` run over #210–#212)

**Accepted — right.** One decision per record is the repo's stated ADR rule (AGENTS.md → Decision records), and the 2026-09-15 audit itself proposed `0028-single-threaded-loopback-http-server` as the home for exactly this. Folding the availability envelope into 0010's note mixed two decisions in one record.

**Fixed in 1aefb49:** the decision now lives in [0028-installer-serve-loop-availability](../blob/fix/p1-server-connection-timeout/docs/decisions/0028-installer-serve-loop-availability.md) (context / decision / consequences, including the 408 semantics and the smoke-only override); 0010's note is a pointer, and the index lists 0028 as the availability half of [0010]. `pnpm check:decisions` clean (28 records).

PR #211, opened 2026-09-15
ci: generate the markdownlint third-party ignores from sync-skill-gates (audit P1)

onemen:
Implements the P1 Skills/Tooling finding from the 2026-09-15 comprehensive audit (the 2026-09-15 working-tree audit (local draft, not committed) §4 / §7). Part of #4 (Phase 5 housekeeping — audit follow-ups).

## The finding

> markdownlint's third-party ignore list is hand-written (`.markdownlint-cli2.jsonc:40-47`) — Generate it from `sync-skill-gates.mjs` and extend the live-repo test to assert all three gates agree.

A new third-party skill currently requires hand-editing the markdownlint config — the exact class of drift the prettierignore managed block already eliminated (the 4dd6640 regression class).

## The change

`sync-skill-gates.mjs` now manages **two** gate files in one pass (`pnpm format:fix`):

- **`.prettierignore`** — unchanged behaviour.
- **`.markdownlint-cli2.jsonc`** — a new managed block inside the `ignores` array holds one `**/.agents/skills/<name>/**` entry per third-party skill, generated from the same `loadInventory()` frontmatter source of truth. Authored skills stay deliberately absent (they ARE linted).

Implementation details:

- **Line surgery, not JSON reserialization** — every comment in the config survives byte-for-byte; prettier keeps the file formatted.
- `findStraySkillLines()` gained optional begin/end marker parameters (the JSONC block uses `//` markers); the tool strips stray hand-written skill entries outside the block in both files on `--fix`.
- Idempotent: `--fix` twice leaves the file unchanged; `--check` fails loudly with per-file line numbers when either block drifts.
- One tool self-trap dogfooded and dodged: a literal `**/` glob inside the tool's own block comments terminates the comment (the file's doc comment warns about exactly this) — doc lines now avoid the sequence.

## Tests (7 new, `sync-skill-gates.test.mjs`)

- Block shape: one entry per third-party skill at the anchor, correct indentation.
- Authored skills absent from the block.
- Stale block replaced + stray hand-written entry stripped (4dd6640 class), unrelated entries preserved, exactly one block after, idempotent.
- Comments and ordering survive byte-for-byte.
- Missing `ignores` anchor fails loudly.
- `findStraySkillLines` with configurable markers.
- **Live-repo sync assertion** — `config/.markdownlint-cli2.jsonc` block matches the current skill set (the audit's "all three gates agree" ask, alongside the existing prettierignore + eslint tests).

## Validation

`pnpm test` 435 pass / 0 fail / 3 skip; `pnpm lint` green (analyzer clean, markdownlint 0 issues; 1 known pre-existing eslint warning); `pnpm format` green; `node tools/sync-skill-gates.mjs` → in sync.


tools/sync-skill-gates.mjs
```diff
@@ -94,7 +105,10 @@ export function classifySkills(root) {
  */
 function isSkillGateLine(line) {
   const t = line.trim();
-  return t !== '' && !t.startsWith('#') && t.includes('.agents/skills');
+  // JSONC `//` comments (markdownlint config prose) are never gate lines —
+  // `--fix` must preserve a comment that merely mentions .agents/skills.
+  if (t === '' || t.startsWith('#') || t.startsWith('//')) return false;
```
onemen, line 110: 🤖 AI review triage (Codebuff agent — result of the batch `review:batch` run over #210–#212)

**Accepted — right.** Verified live before fixing: `renderMarkdownlintConfig` stripped a JSONC prose comment that merely mentions `.agents/skills`, and `findStraySkillLines` flagged it as stray — so a future explanatory comment in the `ignores` array would be silently eaten by `pnpm format:fix` (and `--check` would then fight the file forever).

**Fixed in 1ddf753:** `isSkillGateLine` now treats `//` lines as non-gate lines (same exemption `#` comments already had), and a regression test asserts the comment survives `--fix`, the stray scan stays quiet, and rendering stays idempotent.

PR #214, opened 2026-09-15
fix: replace the admin-copy shell fallback with an in-process copy (audit P2)

onemen:
Implements the P2 Code-quality finding from the 2026-09-15 comprehensive audit (the 2026-09-15 working-tree audit (local draft, not committed) §3/§7). Part of #4 (Phase 5 housekeeping — audit follow-ups).

## The finding

> The Unix installer admin-copy fallback lacks elevation, parent-dir creation, and truncation checks (`admin_copy.c:537-556`) — Replace with the helper/argv-array path or an `execvp` copy; update the `docs/security.md` shell-usage item.

## The change

The `#else` branch (any POSIX platform that is not Linux or Apple — i.e. *BSD and friends) previously did:

```c
snprintf(cmd, sizeof(cmd), "cp \"%s\" \"%s\"", srcs[i], dsts[i]);
system(cmd);
```

Three defects in three lines: argv-interpolated `system()` (the exact pattern `docs/security.md` item 5 forbids — a `"` or `$()` in a detected path breaks the quoting), silent truncation of paths beyond `MAX_PATH_LEN * 3`, and no parent-dir creation.

It now mirrors the Linux branch minus the elevation proxy (these platforms have no pkexec/sudo contract here, by design — unwritable destinations fail with a clear `error_msg`, as before):

- **In-process buffered copy** (`open`/`read`/`write`, full-write loop, `close` errors checked) — no shell, no interpolation, no truncation;
- **`create_parent_dirs()`** — same mkdir-walk as the Linux branch;
- **`errno` in the error message** — a failed copy now says *why* (`Copy failed for %s: %s`), and the elevation-free limitation is documented in a block comment.

`docs/security.md` item 5 updated: the admin-copy fallback is no longer a shell call; the remaining interpolated `system()`/`popen()` calls are enumerated honestly (path-verified `pkill` sweep, `sha256sum` popen on non-ELF platforms, user-dialog helpers).

## Validation — stated honestly

- `gcc -fanalyzer` clean (Windows target, unchanged by this branch), full build + `pnpm test:hash` green.
- The `#else` block compiles on **no** available toolchain as-is (MinGW's `io.h` declares 1-arg `mkdir`; Linux CI builds the `__linux__` branch, macOS CI the `__APPLE__` branch). It was therefore verified **standalone**: the block extracted verbatim and compiled with `-Wall -Wextra -fsyntax-only` against exactly the POSIX headers it includes, with `mkdir` shimmed to its 2-arg POSIX contract — clean, no warnings. (The same shim trick is unnecessary on the real target platforms, where `sys/stat.h` is the only `mkdir` declaration.)
- No runtime test exists for this branch anywhere in CI — it never compiles on the three CI OSes. This is a known, accepted edge (same class as the pre-existing fallback).

No `system(` call remains in `admin_copy.c` (the one grep hit is the comment documenting the removal).


installer/src/admin_copy.c
```diff
@@ -534,14 +545,82 @@ int admin_copy(const char *src, const char *dst, char *error_msg, size_t error_s
...
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+
+static void create_parent_dirs(const char *dst_path) {
+    char dir[MAX_PATH_LEN];
+    snprintf(dir, sizeof(dir), "%s", dst_path);
+    char *p = dir + strlen(dir);
+    while (p > dir && p[-1] != '/') p--;
+    if (p == dir) return; /* filename-only dst: no parent component to create */
```
onemen, line 563: 🤖 AI review triage (Codebuff agent — result of the batch `review:batch` run over #213–#214)

**Accepted — right** (finding 1, minor): verified that a filename-only dst made `create_parent_dirs` call `mkdir("")`. And the scope is wider than reported — the same body exists in the Linux and macOS branches (pre-existing), so the guard now covers all three POSIX twins (this line in each).

**Fixed in `86e7f31`:** early return when no parent component (`p == dir`), plus a root-level guard (`dir[0] == '\0'` — parent is `/`).

installer/src/admin_copy.c
```diff
@@ -534,14 +545,82 @@ int admin_copy(const char *src, const char *dst, char *error_msg, size_t error_s
...
+}
+
+static int copy_file_content(const char *src, const char *dst) {
+    int in_fd = open(src, O_RDONLY);
+    if (in_fd < 0) return -1;
+
+    /* Preserve the source mode (exec bit matters for browser binaries); the
+     * mode is filtered by the process umask as usual. */
+    struct stat src_stat;
+    if (fstat(in_fd, &src_stat) != 0) {
```
onemen, line 587: 🤖 AI review triage (Codebuff agent — result of the batch `review:batch` run over #213–#214)

**Accepted — right** (finding 2, major): hard-coded `0644` dropped the source exec bit — browser binaries copied by the fallback would lose +x. And it was not fallback-only: the Linux branch had the identical hard-code, so both sites (Linux + fallback) are fixed.

**Fixed in `86e7f31`:** `fstat(in_fd, &src_stat)` + `open(..., src_stat.st_mode & 0777)`, umask still applies, mirroring the reviewer's suggested shape.

PR #223, opened 2026-09-16
docs(decisions): ADR 0029 — amendments are first-class status lines, validated

onemen:
## What & why

The 2026-09-15 audit flagged (P2): **amendments changed decisions in place while the index
mandated supersede-don't-edit, and `check-decisions.mjs` could not see them.** An agent reading
only the steering list sees a clean `accepted` record and re-proposes the amended-away behavior.
This PR makes amendments first-class and machine-checked instead of prose-shaped.

Part of #4 (Phase 5 housekeeping) · Audit follow-up · Unowned quick P2 from the audit's §8 table.

## Design (ADR 0029)

An **additive** amendment keeps its record and gets a machine-readable status line under
`- **Status:**`; a **reversal** still supersedes (unchanged):

```markdown
- **Amends:** [0019](./0019-release-versioning.md) (what changed)   # on the amending record
- **Amended:** [0029](./0029-status-line-amendments.md) (what changed)  # on the amended record
```

A date-only `- **Amended:** YYYY-MM-DD — what changed` line (no link) is also valid — declared,
not validated. Supersede remains the tool for reversing; status lines record additive amendments
without renumbering history.

## Checker changes (`tools/check-decisions.mjs`)

- **New ADR 0029 validation** — an `Amends:`/`Amended:` link must point at an existing record,
  must not be the index/template or a self-link, its `[NNNN]` must match the target filename, and
  the pair must be **reciprocated** (X Amends Y ⇔ Y Amended X). Reciprocity is checked from both
  ends so a one-sided edit fails whichever record it landed on. Repeated status keys (two
  `Amends:` lines) accumulate; a continuation line that itself starts with `**Key:**` is a new
  field, never a join — adjacent `Amends:`/`Amended:` items keep their links separate.
- **Link scanning** now ignores fenced code blocks (format examples) and inline code spans.
- **Testability refactor** — the whole check is a pure exported `checkDecisionsDir(dir)` plus
  `parseStatusBlock(lines)`; the CLI is behind a main guard. 12 new unit tests in
  `test/unit/tools/check-decisions.test.mjs` drive fixture directories — the real decision log
  stays validated by `pnpm check:decisions` itself (CI).

The working tree also demonstrated the point twice while authoring: the checker caught a missing
newline fusing two status fields in 0026, and a one-sided `Amended:` link on 0012.

## Existing amendments formalized

- **0019** — its 2026-09-12 and 2026-09-13 amendment sections now carry `Amended: [0029]`.
[...]

PR #224, opened 2026-09-16
test(publish): generated-file registry ↔ hash-inputs consistency gate (ADR 0008)

onemen:
## What & why

The 2026-09-15 audit flagged (P2): **`syncGeneratedFiles.mjs`'s generated-file registry and the
publish hash inputs were parallel hand-maintained lists** — exactly ADR 0008's stated trap
(*"adding/removing a generated file without updating the hash inputs silently stops updates
propagating"*). The `gitignoreUtils` half of the finding shipped in #207; this PR is the registry ↔
hash-inputs half.

Part of #4 (Phase 5 housekeeping) · Audit follow-up · Unowned quick P2 from the audit's §8 table.

## Design: one registry, every consumer derived

New **`tools/publish/generatedRegistry.mjs`** — the single authoritative list of the four generated
files (ADR 0008) with how each enters the hashes and zips:

- `GENERATED_FILES` — rel, generator provenance, and `shipsIn` (which package's zip carries it, at
  which rel; the installer headers ship nowhere — their sources are hashed instead)
- `PACKAGE_SCAN_EXCLUDE` / `INSTALLER_HASH_EXCLUDE` / `OBSOLETE_FILES` — the scan-drop and
  installer-hash exclusion lists
- `PACKAGE_ROOTS` + `packageExtraFiles(name)` — the shared `{rel, absPath}` entries

`packageExtraFiles()` output is passed **verbatim** to both `computeDirectoryHash` (hash + manifest
`files` list) and the `createZip` re-add, so the zip content and the canonical list can no longer
diverge even in principle.

**`syncGeneratedFiles.mjs`** derives `GENERATED`/`PREVIEW` from the registry (only the generator
functions live there); output verified byte-identical by the new test's `regenerate()` idempotence
assertion and by a full `upload:local` snapshot.

**`upload.mjs`** loses the hand literals (`GENERATED_UPDATER_CONFIG`, `GENERATED_UI_CSS`, the
inline `['_config.h', 'resources.h']` exclude, `HASH_EXCLUDE = ['versionInfo.json']`, the
hand-built package roots) — all registry calls now. Adding/removing a generated file is a
**one-place edit**: the registry, and the tests tell you what must follow.

## The test (`test/unit/generatedRegistry.test.mjs`, 10 assertions)

Registry well-formedness (unique rels, sources on disk, all gitignored per ADR 0008, every file has
a generator); `extraFiles ⊆ registry shipping rels` per package; build products ⊆ installer-hash
excludes; obsolete files still excluded from zip + hash pipelines; and cross-checks that read the
**live** `upload.mjs`/`syncGeneratedFiles.mjs` sources so a hand-literal regression (or a generator
dropped from the registry) fails `pnpm test`.

## Docs sync

ADR 0008 gets an `Amended: 2026-09-16` status line (per ADR 0030's date-only form — #223) plus a
Consequences note: the trap is now closed mechanically. The `generated-files` skill, `AGENTS.md`
architecture invariant, and the `DEVELOPING.md` tool tree name the registry.

## Validation

- `pnpm test` — 451 pass / 0 fail (10 new)
- `pnpm lint` (eslint + tsc + C format + gcc -fanalyzer + markdownlint) — clean
- `pnpm format` — clean
- `pnpm upload:local -- --mode=prod` — full snapshot built; manifest `files` lists identical in
  shape to pre-refactor (`updater/updater-config.sys.mjs` in utils, `updater.css` in updater-ui)
- `pnpm test:hash` — JS ↔ C hash parity PASS (utils, fx-folder, adversarial comparator)

🤖 Generated with Codebuff


tools/publish/syncGeneratedFiles.mjs
```diff
@@ -110,42 +111,40 @@ export function buildRemoteUiCss() {
...
 // publish time so it ships inside updater-ui.zip.  Never tracked or committed:
 // it is regenerated on demand from the shared design system + updater tail.
-const PREVIEW = [
-  {
-    rel: 'tools/publish/remote-ui/updater.css',
-    generate: buildRemoteUiCss,
-  },
-];
+// Also registry-driven (the one PREVIEW member).
+const PREVIEW = GENERATED_FILES.filter(f => !(f.rel in GENERATORS)).map(f => ({
```
onemen: 🤖 **AI review — agent-posted (ADR 0020; Codebuff agent — result of the local review:local run)** — minor — accepted as **right**: `PREVIEW` mapped any registry file outside `GENERATORS` to `generate: null`, so a future registry entry without a generator would surface only as a `TypeError` inside `regenerate()`. Fail-fast convention says a load-time error naming the file. **Fixed in 38d5c6f** (this head commit).

onemen (commented):
🤖 AI review — agent-posted (ADR 0020; Codebuff agent — result of the local `review:local` run on p2/gen-registry-hash-inputs)

Provider/model: openrouter (per-file fallback; gemini rate-limited). 8 files in the diff; 7 reviewed, 1 skipped on transient provider errors.

**Findings: 1 accepted (right), posted as its own line-anchored thread.**

- tools/publish/syncGeneratedFiles.mjs:141 — minor — `PREVIEW` mapped a registry file without a generator to `generate: null` (a future entry would die as a TypeError inside `regenerate()`). Fixed in 38d5c6f: a load-time error names the file and the fix.

**Assessed wrong/useless: 0.**

Coverage caveat (honest-zero rule): `tools/publish/generatedRegistry.mjs` was skipped in this pass on transient provider errors. Compensating checks: the module is constants-only (verified by reading it in full), its correctness is pinned by the new test suite (10 assertions, including regenerate() idempotence), and `pnpm upload:local -- --mode=prod` + `pnpm test:hash` prove the derived hash inputs byte-match the pre-refactor behavior. Not re-running the reviewer on one file at the cost of more skipped-file rotation; flagged here per protocol.

Advisory only — this review never blocks the merge (ADR 0020).

PR #226, opened 2026-09-16
feat(lint): check-skills gate — SKILL.md frontmatter validation + vendored skill tests (audit P2)

onemen:
Closes the 2026-09-15 audit's last unowned P2 (§8 "Still open"): **SKILL.md frontmatter validation + running the vendored skill's tests** — "`skills-watchdog.mjs` parses frontmatter but nothing gates on it".

Fixes the finding by adding `tools/check-skills.mjs`, wired into `pnpm lint` (static-only alias: `pnpm test:skills`).

## What it checks

Per skill directory under `.agents/skills/`:

- **SKILL.md exists** — a directory without one is invisible to every other tool; flagged, not silently skipped.
- **Flat tree (ADR 0022 decision 4)** — every SKILL.md must sit at `.agents/skills/<name>/SKILL.md`; a nested SKILL.md means a second discovery root. Vendored internal layout (`scripts/`, `references/`) is fine — the rule bounds the roots, not their contents.
- **`name` present, non-empty, = directory name**; **`description` present and non-empty** (folded multi-line blocks count).
- **Gh metadata is all-or-nothing**: third-party skills (`metadata.github-repo`) must carry all four gh-injected keys with a parseable repo URL; an *authored* skill carrying partial `metadata.github-*` keys is flagged as the manually-copied-skill signature ADR 0022 retired.

Then every vendored `*.test.mjs` runs under `node --test` (today: `debugging-firefox/scripts/firefox-rdp.test.mjs`, a pure-Node suite — no Firefox needed).

## Design notes

- **Validation, not mutation** — ADR 0022: "gates that mutate apply to things we author; gates that validate apply to everything." No vendor text is linted or formatted; the eslint/prettier/markdownlint ignore story is untouched.
- **One parser** — frontmatter parsing reuses `parseSkillFrontmatter`/`loadInventory` from `tools/skills-watchdog.mjs`; no duplicate classification logic.
- **14 unit tests** (`test/unit/tools/check-skills.test.mjs`) pin the rules with fixture directories, mirroring the checker-test conventions of `check-decisions`.

## Docs sync

AGENTS.md (lint command + skills bullet), `docs/ci-inventory.md` (lint row), `docs/DEVELOPING.md` (skills checker bullet), `package.json` (`lint`, `test:skills`).

## Validation

- `pnpm test` — 479 pass / 0 fail (14 new)
- `pnpm lint` — green, ending with `check-skills: OK (1 vendored test file(s) run)`
- `pnpm format` — green

PR #228, opened 2026-09-16
feat(security): CSP meta on the installer page and updater tab (audit P3)

onemen:
## Summary

Closes the audit's P3: neither shipped HTML surface declared a Content-Security-Policy, so both ran under the browser default. Part of the 2026-09-15 audit follow-up queue (F4), sequenced after #226/#227 per owner call.

- **Installer page** (`installer/web/index.html`, served by the C server on `127.0.0.1:8777`): strictly same-origin policy — `script-src 'self'` (`/script.js`), `style-src 'self'` (`/style.css`), `img-src 'self'` (`/logos/*`), `connect-src 'self'` (`/api/*` under the ADR 0010 session-token model). No inline script or handlers exist today; the policy keeps it that way.
- **Updater tab** (`tools/publish/remote-ui/updater.html`, chrome-privileged): content locked to its own `chrome://firefox-scripts` package (updater.js, updater-ui.js, generated updater.css, logos); `default-src 'none'` baseline and `connect-src 'none'` — package/helper/installer downloads go through the privileged `Downloads` API, not the document, so no remote hosts are needed.
- Both: `object-src 'none'`, `base-uri 'none'`, `form-action 'none'`, `frame-ancestors 'none'`, and no `unsafe-inline`/`unsafe-eval` anywhere.
- `test/unit/publish/cspMeta.test.mjs` pins both policies (directive-by-directive) and fails `pnpm test` if an inline script/handler appears or a policy loosens.

## Validation

- `pnpm test` — 488 tests, 0 failures (4 new CSP assertions)
- `node tools/publish/syncGeneratedFiles.mjs --check` — in sync (regenerated `resources.h` embeds the new installer HTML)
- `pnpm lint` (now strict, `--max-warnings 0`) and `pnpm format` — clean

🤖 Generated with Codebuff


installer/web/index.html
```diff
@@ -3,6 +3,13 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <!-- CSP: strictly same-origin — /script.js, /style.css, /logos/*, /api/*
+         (session-token model, ADR 0010). No inline script/handlers.
+         frame-ancestors is meta-ignored by spec; a header would need a C-server change (follow-up). -->
+    <meta
```
onemen, line 9: 🤖 **AI review triage (Codebuff agent — result of the local `tools/ai-review.mjs` run)** — accepted as **right**: `frame-ancestors` is ignored by browsers when delivered via `<meta>` (CSP3 delivery rules), so the directive implied false protection. Fixed in a7edf0a: removed from both metas, tests now assert its absence, and the comment notes an HTTP response header on the C server as the follow-up if framing hardening is wanted. (The two other review files were provider-skipped; the test file is the gate that pins this whole policy and is covered by the compensating manual review — directives enumerated against the pages' actual load surfaces.)

PR #232, opened 2026-09-17
refactor(installer): extract the restart suite from main.c into restart.c (audit F5)

onemen:
Part of the 2026-09-15 audit's §3.1 modularity follow-ups (F5) — the first of the one-split-at-a-time series (§7 deferred row: "Split one at a time per §3.1").

## What

`installer/src/main.c` was **2855 lines** mixing HTTP handlers, restart orchestration, browser scan, CLI, and self-update. This PR carves out the **restart suite** (browser close / graceful relaunch / session bookkeeping / the Windows async restart worker) verbatim into **`restart.c` (526 lines) + `restart.h` (74)**, leaving **main.c at 2348**. The HTTP handlers in main.c decide *WHEN* to restart; the new module owns *HOW*.

## Seams (zero behavior change)

- **Public API** (`restart.h`): `close_browser_by_pid`, `close_browser_binary` (also used by `handle_api_close_browser`), `open_url_in_profile` (used by `handle_api_open_folder`), `do_restart_work`.
- **Async worker seam**: the pristine main.c spawn block (single-flight guard → `CreateThread` with a heap-copied plan → on failure reset flag + free plan → synchronous fallback) becomes `restart_start_async()` / `restart_worker_busy()` under `#ifdef _WIN32`, with the guard state now internal to restart.c. The `handle_api_restart` diff is 21 lines → 1 call.
- **Shared state**: `verbose_printf` moves to `platform.h` over a shared `installer_verbose_flag`; 6 shared state definitions in main.c drop `static` with `extern` declarations in `restart.h`.
- **Build**: no Makefile change — `src/*.c` wildcard compiles restart.c automatically; POSIX/macOS/BSD paths compiled out via existing `#ifdef` structure (link-validated by CI's cross-OS matrix).

## Validation

- `make dist_win` links clean (MSYS2 UCRT64); artifact verified (`MZ` magic, `strings` confirms the restart code paths present).
- `make analyze`: gcc -fanalyzer — 0 findings across 7 files.
- `pnpm lint` (incl. the new strncpy gate), `pnpm format`, `pnpm test` — 490 pass / 0 fail.
- Extraction was script-driven with per-symbol assertions; the moved block is byte-identical to main.c@`3684f24` lines 1697–2184 modulo the documented seam edits above.

## Not in this PR

The other §3.1 splits (`e2e.yml`, `check-browser-downloads.mjs`, `remote-ui/script.js`) — one at a time per the audit's own prescription; `check-browser-downloads.mjs` needs the ADR 0025 path-filter context first.

🤖 Generated with Codebuff


onemen:

**Three `review:local` passes** (free-tier provider flakiness shuffled coverage between runs):
- Pass 1: 0 findings; `restart.c` skipped (transient provider unavailability).
- Pass 2: 0 findings; `main.c` + `restart.c` skipped, one invalid JSON reply on `restart.h`.
- Pass 3: 2 reported on main.c/restart.h — the **restart.h** note found no issues; the **main.c** finding ("restart_worker_thread calls restart_start_async which is undefined; entire restart module removed → compilation error") was triaged as **wrong**: the call direction is the opposite (`restart_start_async` spawns `restart_worker_thread`, and the worker body only calls `do_restart_work`), the module exists in the new `restart.c`/`restart.h` (the diff-only view missed the added files), and the claim is disproven by the green `make dist_win` link in the PR description.

**`restart.c` never got a completed provider review** (transient skip ×3). Compensating evidence: the file is ~90% code moved byte-identical from main.c@`3684f24` (script-asserted), gcc -fanalyzer runs over it with 0 findings, `dist_win` links, and it passes the format/strncpy gates. Nothing accepted; no threads opened.

installer/src/restart.c
```diff
@@ -0,0 +1,547 @@
...
+    return 0;
+#else
+    char launch[MAX_PATH_LEN];
+    snap_launcher_path(b->binary_path, launch, sizeof(launch));
+    const char *tab = (url && url[0]) ? url : NULL;  // execl's argv ends at
+    // the first NULL, so a NULL url must drop --new-tab entirely, not pass a
+    // dangling switch (same guard as the Windows branch above).
+    pid_t child = fork();
+    if (child == 0) {
+        setsid();
```
onemen, line 326: 🤖 Accepted (CodeRabbit review:batch, triaged by the Codebuff agent): verified **right**, fixed in c3b9a04.

The Windows branch already guarded `url && url[0]` but the POSIX fork/execl branch passed `url` straight through; since execl's variadic argv list ends at the first NULL pointer, every non-UI-host profile relaunch on Linux/macOS got a dangling `--new-tab` with no argument. The fix mirrors the Windows guard: without a URL, `--new-tab` is dropped entirely (both the with-profile and no-profile execl calls).

Note: pre-existing in the moved code (pristine main.c had it too) — the split didn't introduce it, but the review caught it.

installer/src/main.c
```diff
@@ -30,7 +32,7 @@
 #include <stdbool.h>
 
 // ===== Verbose logging =====
-static int g_verbose = 0;
+int installer_verbose_flag = 0;  // shared via platform.h (restart.c reads it)
```
onemen, line 35: 🤖 Accepted (CodeRabbit review:batch, triaged by the Codebuff agent): verified **right** — and this one was a genuine split regression, fixed in c3b9a04.

The split left the file-local `static int g_verbose` in place while `verbose_printf` moved to platform.h over `installer_verbose_flag`; the parser set `g_verbose` at all four sites (--verbose, --smoke-test, --server-only, --scan-only) but the macro read the never-assigned shared flag, so all verbose output went silent. Fix: `g_verbose` is gone entirely, the four parser sites now set `installer_verbose_flag` — one real variable, no macro aliasing.

Issue #233, opened 2026-09-17
Dev environment: Defender real-time scan intermittently breaks local Windows links (collect2: ld returned 5, truncated exe)


onemen:
Part of #38 (Post v1.0). Observed repeatedly during the restart-split work (PR #232 session, 2026-09-17).

## Symptom

On a local Windows host (MSYS2 UCRT64, binutils ld 2.47), an ad-hoc or Makefile-driven link of `installer_win.exe` intermittently fails:

```
collect2.exe: error: ld returned 5 exit status
```

Exit code 5 = `ERROR_ACCESS_DENIED`. The failure is *silent* — no undefined-reference or other diagnostics — and the output file is either:

- present but **truncated/corrupt** (file starts `00 00` instead of `MZ`), or
- **vanishes** between commands (observed once: listed in one command, gone in the next).

Retrying the same link usually succeeds; object-file-level links succeeded 3/3 while fresh one-shot multi-TU links (~620 KB unstripped output) failed several times in a row. The official `make dist_win` failed once, then passed on the next invocation with zero code changes.

## Diagnosis

This is **not** the #157 ML-detection issue — `Get-MpThreatDetection` shows no new detections and the Defender event log is clean. The pattern matches **Defender real-time protection holding a write lock on the freshly written unsigned .exe**: the scanner opens the file for scanning as `ld` is finishing its write/stamp pass, and the subsequent operation fails with access denied. Longer/multiple-TU outputs (bigger write window) correlate with more failures. CI runners are unaffected (Linux/macOS legs, and GitHub's Windows runners don't exhibit it).

## Workaround recipe (what worked during #232)

1. **Just retry** — a second identical link command succeeds in the vast majority of cases (Defender finishes its scan and releases the lock).
2. Prefer `mingw32-make -C installer dist_win` over ad-hoc `cc file1.c file2.c …` one-shot links — the Makefile's single link step is easy to re-run, and a corrupt partial exe is obvious from the `ld` failure.
3. **Verify the artifact after a retry**: `head -c 2 dist/installer/installer_win.exe` must print `MZ`; `file dist/installer/installer_win.exe` must say `PE32+ executable`. Never ship or hash-test a binary you only saw "exit 0" for if a previous run in the same session hit exit 5.
4. If a retry loop is desired: `for i in 1 2 3; do make ... && [ "$(head -c 2 out.exe)" = "MZ" ] && break; sleep 2; done`.

## Candidate follow-ups (to decide when picked up)

[...]

PR #234, opened 2026-09-17
feat(publish): magic-byte artifact check before hashing/publishing (issue #233)

onemen:
Closes the first candidate follow-up of #233 (Defender write-lock race on local Windows links).

## What

`tools/publish/upload.mjs` now verifies every staged installer/helper binary starts with its platform's executable magic before anything downstream consumes it:

- **After the build/reuse pass** — catches a truncated artifact immediately after `make`, naming the file and the remedy (delete + rebuild; the link usually succeeds on retry).
- **In pass 2 (`--skip-build`)** — the SignPath round-trip moves binaries across machines/artifact transport, so the pass that runs the AV/VT gates and publishes re-verifies what it is about to gate.

## Shape

`verifyStagedBinaries(files, access)` lives in `tools/publish/platforms.mjs` as a **pure function** (injected name→Buffer accessor — consistent with the module's dependency-free registry style and directly unit-testable): PE `MZ` for win, ELF for linux/aarch64, all four Mach-O magics for mac. Bad names are collected and reported together.

## Validation

- 7 new unit tests (magics per platform, Mach-O variants, the #233 `00 00` truncation, empty/missing, wrong-magic, aggregate reporting, short-buffer edge) — full suite 500/0.
- `pnpm lint` / `pnpm format` green; docs sync: DEVELOPING.md AV-hardening section gains the third gate entry.

Refs #233 (Part of #38).


tools/publish/platforms.mjs
```diff
@@ -83,3 +81,46 @@ export function helperShaAssetName(p, suffix = '') {
...
+const MACHO_EXTRA = [
+  [0xce, 0xfa, 0xed, 0xfe], // MH_MAGIC / MH_CIGAM (32-bit arch)
+  [0xca, 0xfe, 0xba, 0xbe], // FAT_MAGIC / FAT_CIGAM (universal wrapper)
+  [0xca, 0xfe, 0xba, 0xbf], // FAT_MAGIC_64 / FAT_CIGAM_64
+];
+
+export function verifyStagedBinaries(files, access) {
+  const bad = [];
+  for (const [name, p] of Object.entries(files)) {
+    const buf = access(name);
```
onemen, line 118: 🤖 Accepted (review:local finding, triaged by the Codebuff agent): verified **right**, fixed in the follow-up commit.

Verified against Apple cctools `mach-o/fat.h`: `FAT_MAGIC` is `0xcafebabe` and `FAT_MAGIC_64` is `0xcafebabf` — in file order `ca fe ba be` / `ca fe ba bf`, not the byte sequences this check originally listed (fat-header values misread as file bytes). A macOS universal binary would have been flagged invalid. Also confirmed from `mach-o/loader.h`: the cigam variants are byte-identical to their magic twins (they differ only in the reader's byte-order interpretation), so the accepted set is otherwise unchanged. Unit test updated to pin all four accepted mac magics with the cctools reference.

tools/publish/platforms.mjs
```diff
@@ -83,3 +81,72 @@ export function helperShaAssetName(p, suffix = '') {
...
+
+/**
+ * Structural check behind verifyStagedBinaries: the platform's header magic
+ * must match AND the file must be long enough to be more than a header — a bare
+ * `MZ` (or any header-only stub) is exactly the truncation the check exists to
+ * catch.
+ */
+function looksLikeExecutable(buf, p) {
+  const magic = MAGIC[p];
+  if (!magic) return true; // unknown platform — naming tests cover the registry
```
onemen, line 132: 🤖 Accepted (CodeRabbit review:batch, triaged by the Codebuff agent): verified **right**, fixed in this commit.

A bare `MZ` buffer passed the original magic-only check. `looksLikeExecutable` now enforces structure, not just a prefix: for PE, the DOS header's `e_lfanew` (offset 0x3c, little-endian) must point at a `PE\0\0` signature inside the file; all platforms enforce a minimum length above the header floor (real installers/helpers are hundreds of KB). Tests added for the exact suggested case (two-byte MZ), MZ+padding without a signature, valid PEs at several `e_lfanew` offsets, and `e_lfanew` pointing past EOF.

tools/publish/upload.mjs
```diff
@@ -445,9 +446,16 @@ async function buildBinaries(platforms, storedHashes) {
     // dev). A genuinely idle prod pass (nothing changed, nothing staged) still
     // completes — it uploads nothing and leaves the manifest untouched.
     const missing = [];
+    const stagedPass2 = {};
     for (const p of platforms) {
-      if (installerChanged && !fs.existsSync(installerPath(p))) missing.push(installerAssetName(p));
-      if (helperChanged && !fs.existsSync(helperPath(p))) missing.push(helperAssetName(p));
+      if (installerChanged) {
+        if (!fs.existsSync(installerPath(p))) missing.push(installerAssetName(p, ASSET_SUFFIX));
+        else stagedPass2[installerAssetName(p, ASSET_SUFFIX)] = p;
```
onemen: 🤖 Accepted (CodeRabbit review:batch 2, triaged by the Codebuff agent): verified **right**, fixed in this commit.

Same root cause as the build-pass suffix bug the publish gate caught: pass 2's `stagedPass2` keys (and the missing-artifact error list) called the asset-name helpers without `ASSET_SUFFIX`, so a dev-mode pass 2 would read the unsuffixed path the existence check never wrote — `ENOENT` on the file it just verified exists, or a stale-artifact validation. Both maps now use the suffixed names, matching `installerPath()`/`helperPath()`.

onemen:
🤖 AI review triage (Codebuff agent — result of the local ADR 0020 review rerun before merge): 0 findings across all 4 files (platforms.mjs, upload.mjs, DEVELOPING.md, test). The three earlier threads (verbose-flag alias, Mach-O fat magics, pass-2 suffix) are all resolved with fixes in-PR.

PR #235, opened 2026-09-17
test(e2e): surface per-poll failures in pollUntil instead of opaque timeouts

onemen:
Follow-up from the RS-10 investigation (the Windows E2E flake observed during PR #232's CI).

## Problem

RS-10 polls `/api/status` for up to 45s. In the flaked run, the whole budget elapsed with **zero diagnostic output**: the poll loop's `catch { /* transient */ }` swallowed every fetch failure, so a server that stopped answering was invisible until the check failed with an empty detail. The installer-side log was equally silent — undiagnosable after the fact.

## Change

- **`pollUntil` moves to `test/e2e/shared/helpers.mjs`** with per-attempt error visibility: a thrown error is logged throttled (first attempt, then every 10th), and a timeout with a last error prints it. Same truthy/null contract as before.
- **All six installer-e2e call sites get labels** (error-lines-only, zero happy-path noise).
- **The RS-10 status poll logs each failed fetch inline** — that site's errors *are* the diagnostic (server not answering vs. state machine stuck vs. JSON parse).
- Deliberately **not** changed: the 45s budget (the successful run completes the same step in ~0.35s; raising the ceiling would only slow failures — the investigation showed observability, not time, was the gap).

## Validation

- `node --check` both files; `pnpm lint` / `pnpm format` / `pnpm test` (490 pass) green.
- Happy-path behavior identical; error paths now produce log lines.

Related: #233 (Defender write-lock issue, same CI investigation), umbrella #38's existing E2E-observability items.


onemen:

First `review:local` pass skipped both files (transient provider unavailability); second pass reviewed both — **0 findings**.

onemen:
🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run): 1 major — `pollUntil` could hang forever if a poll callback never resolved (e.g. a fetch without its own timeout), extending the run past the budget it advertises. **Accepted** — fixed in d074051: each attempt races the remaining deadline and the inter-attempt sleep is capped at it, so the poll always returns on schedule. Nitpicks: none material.

onemen:
🤖 AI review triage (Codebuff agent — result of the local ADR 0020 review rerun before merge): 0 findings across both files (shared/helpers.mjs, installer-e2e.mjs). The CodeRabbit batch major (pollUntil could hang past its budget on a never-resolving callback) is accepted and fixed in d074051 — see the earlier triage comment.

Issue #132, opened 2026-09-05
macOS universal installer binary (arm64 + x86_64)


onemen:
Part of #38 (Post-v1.0 roadmap). Source: scope audit absorbed into `docs/roadmap.md` + `docs/future-work.md` §6 (PR #128).

`dist_mac` currently builds a native-arch-only installer; the helper is already universal. Decide whether the installer must be universal (arm64 + x86_64 in one binary) or per-arch downloads suffice.

Notes:

- Interplay with ADR 0019 (stable unversioned artifact names — a single universal `installer_mac` asset vs per-arch names).
- Build cost on CI runners (universal needs both `-arch` slices or two builds + `lipo`).

PR #236, opened 2026-09-17
ci(e2e): extract setup-browser / build-installer / linux-test-prereqs composites (audit F5 §3.1)

onemen:
Part of the 2026-09-15 audit's §3.1 modularity follow-ups (F5) — the e2e.yml split from the deferred §7 row ("Split one at a time per §3.1, keeping pnpm check:gates green after the workflow split"). Follows the restart-suite split (#232).

## The duplication

The 13-job e2e.yml (1264 lines) carried four duplicated step groups:

1. **Browser download-URL resolve + cache + install** — five jobs (installer, updater, browser-matrix, updater-waterfox, fork-portable), each with its own cache-key prefix and redirect-hop variant (single-hop vs two-hop URL resolution, bash vs pwsh twins).
2. **Native-installer build block** with the #132 universal-mac `lipo` assertion and the INSTALLER_BIN export — twice, byte-identical (portable-firefox, snap-firefox).
3. **harden-apt + xvfb prerequisites** — five jobs.
4. **Portable-dir setup** — three jobs, bash + pwsh twins.

Every recipe fix had to be applied N times — the drift the audit's modularity section flags.

## The split

Three composite actions become the single copy (`.github/actions/`):

| Action | Absorbs | Parameterization |
| :--- | :--- | :--- |
| `setup-browser` | URL resolve → installer cache → portable dir → extracted-dir cache → `downloads.mjs` install | `browser`, `cache-key-prefix` (preserves each job's key prefix: `firefox-dl` / `firefox-portable` / `browser-dl`), `portable` |
| `build-installer` | case-over-runner.os `make dist_*` + #132 lipo assertion + `INSTALLER_BIN` export (incl. the Windows `github.workspace` path caveat) | — |
| `linux-test-prereqs` | harden-apt + xvfb install | — |

Deliberately **not** extracted: the changes/paths-filter job and the e2e-gate wiring (the gate-coverage contract parses those shapes directly), and the snap-firefox snapd probe (single-site logic).

## Contract safety

- `pnpm check:gates` passes **unchanged** — job names, `needs:`, path-filter `if:`s, timeouts, matrices, and the gate's applicability/results wiring are byte-preserved; the path filters already list `.github/actions/**`, so the new composites are inside every affected group's filter.
- `docs/ci-inventory.md` needs no row changes (no workflow/job name, trigger, or gate semantics changed) — the Filter ownership table's `.github/actions/**` entries already cover the new files.
- `pnpm format` (prettier on YAML) green; e2e.yml drops 1264 → ~1030 lines.

## Validation honesty

Local validation covers YAML shape (prettier), the gate-coverage contract, and a job-by-job diff review against the pre-split file. The composite steps' runtime behavior (cache key identity, `FIREFOX_BINARY`/`INSTALLER_BIN` exports) is validated by the full E2E matrix on this PR — that is what this PR's own CI run exercises.

Refs #38, part of the audit follow-up series (#232, #234, #235).


onemen:

Two `review:local` passes (free-tier flakiness shuffled coverage between runs):
- Pass 1: 0 findings; `setup-browser/action.yml` skipped (transient provider unavailability).
- Pass 2: 0 definite findings; the `setup-browser` note raised three non-definite observations, triaged here:
  - *Portable-dir cache misses when the URL changes but extraction doesn't*: intentional — the `-x` key follows the same URL hash as the installer cache so a version bump can never restore a stale extracted browser under the new key (the pre-split portable-firefox job documented this exact invariant; the extracted Firefox must match the resolved version).
  - *Windows portable step not error-checking directory creation*: pre-existing pattern carried over verbatim from the jobs (New-Item -Force is idempotent and non-throwing in this context); left as-is to keep the split behavior-preserving.
  - *BROWSER_DL_DIR under runner.temp cleaned between steps*: runner.temp persists for the whole job lifetime, not per-step — non-issue.
- One caveat for reviewers, noted in the PR body too: the composite steps' runtime behavior (cache key identity, FIREFOX_BINARY/INSTALLER_BIN exports) is exercised by this PR's own E2E matrix run — a green run here is the runtime validation.

onemen:
🤖 AI review triage (Codebuff agent — result of the local ADR 0020 review, two passes for full coverage): 1 finding — the mac universal check executed both lipo-thinned slices, which fails on an x86_64 runner (arm64 slice can't execute; set -e aborts). **Accepted** — fixed in 786fbe9: the host-arch slice runs, the other is asserted structurally via lipo -info. The block was carried verbatim from e2e.yml's portable-firefox job; the split surfaced it. Pass 2 covered the remaining files: 0 findings.

onemen:
🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run: one combined run over #234 + #235 + #236, 1 finding across the 10-file merged diff): minor — the `setup-browser` composite's `portable-dir` output referenced `steps.portable-dir`, an id no step had, so the declared output could never resolve. **Accepted** — fixed in 4df6b37: distinct ids on the two platform steps (portable-dir-unix / portable-dir-windows), coalesced in the output value. No caller reads the output today (both portable legs use the `PORTABLE_BROWSER_DIR` env export), so this was latent. All other files clean.

PR #237, opened 2026-09-17
refactor(watchdog): extract the pure reporting layer into tools/ci/watchdog-report.mjs

onemen:
Third of the §3.1 modularity splits (audit the 2026-09-15 working-tree audit (local draft, not committed) P3 row — after #232 main.c/restart.c and #236 e2e.yml composites).

## What

`tools/check-browser-downloads.mjs` (1313 lines) mixed three concerns: vendor network resolution, baseline/drift persistence, and GitHub-visible rendering. This PR moves the **pure layer** into a new `tools/ci/watchdog-report.mjs` (~560 lines, zero imports, zero I/O):

- Domain constants: `BROWSERS`, `VALIDATED_BROWSERS`, `FORK_BROWSERS` (ADR 0021/0025 tiering), `WATCHDOG_LABEL`, `META_ISSUE_TITLE`, `HISTORY_PER_BROWSER`
- Drift classification: `compareBaseline`, `collectDrift`, `collectValidatedDrift`
- E2E dispatch planner: `planDispatches`
- Rendering: `buildStatusTable`, `updateHistory`, `seedHistoryFromBaseline`, `renderHistory`, `buildMetaIssueBody`, `issueTitle`, `issueBody`, `isFailureIssueTitle` + all formatters and `escapeTableCell`

The watchdog keeps network resolution, baseline persistence, GitHub API calls and CLI orchestration, and **re-exports the moved names** — `test/unit/tools/check-browser-downloads.test.mjs` (687 lines) and `tools/ci/record-validated-versions.mjs` are untouched.

## Deliberately unchanged

- Behavior: the moved code is verbatim; the orchestration steps are byte-equivalent.
- Consumers: no import changes anywhere.
- CI contract: job names, `needs:` and gate wiring untouched — `pnpm check:gates` passes.
- The `browser-downloads` filter-ownership row in `docs/ci-inventory.md` already covers "watchdog tooling"; the url-watchdog PR path allowlist gains the new file so tool-layer changes keep their stateless `--pr` smoke run.

## Validation

- `pnpm test` — 504 tests, 0 fail (the watchdog suite imports through the re-export surface, unchanged)
- `pnpm lint` (incl. `tsc` checkJs over `tools/`) and `pnpm format` — green
- `pnpm check:gates` — green
- CI's stateless `--pr` watchdog leg exercises the CLI live on this PR

Part of the Post v1.0 modularity series; remaining §3.1 item: `installer/web/script.js`.

PR #238, opened 2026-09-17
refactor(installer-web): split script.js into phase parts concatenated at build time

onemen:
Final §3.1 modularity split (audit the 2026-09-15 working-tree audit (local draft, not committed) P3 row — after #232 `restart.c`, #236 e2e.yml composites, #237 watchdog-report).

## What

`installer/web/script.js` (1930 lines, one IIFE) mixed debug UI, state, ingestion, banners, rendering, installation and bootstrap. It is now authored as **six phase part files** under `installer/web/script/`, concatenated in order by `installer/embed.mjs`:

| Part | Phase | Lines |
| --- | --- | --- |
| `00-head.js` | IIFE open, debug banner, logos, state | 109 |
| `10-ingest.js` | utilities, grouping, API/network helpers, remote-data ingestion | 367 |
| `20-banners.js` | self-update banner (ADR 0019), test/dev build banner | 111 |
| `30-render.js` | browser list rendering | 431 |
| `40-install.js` | group installation, per-card progress, shutdown helpers | 590 |
| `50-init.js` | init/bootstrap, IIFE close | 322 |

## Why concat, not modules

The C installer embeds **one** gzipped `RES_SCRIPT_JS_GZ` and `index.html` loads **one** `<script>` — introducing ES modules would change the served asset contract (and possibly the C resource plumbing). Concatenation keeps the served artifact byte-compatible in shape: every part shares the single IIFE scope, `00-head` opens it, `50-init` closes it. Zero C changes.

## Safety

- `embed.mjs` **syntax-checks the exact concatenation** (`node:vm` `Script`) on every embed — the Makefile `resources` target runs on every build and `createZip.mjs` at publish, so no gate can ship a broken asset.
- The Makefile `resources` rule tracks `$(wildcard web/script/*.js)` instead of the monolith — a part change regenerates.
- eslint + prettier skip the fragments (not parseable standalone), via documented ignores in `config/eslint.config.js` / `config/.prettierignore` — the same carve-out pattern as `core/fx-folder`. The parts are verbatim-extracted: a normalized diff against the pristine monolith is **empty**; the gzipped served asset is 20,236 bytes vs 20,232 pristine (the seam newlines only).

## Validation

- `dist_win` links (`PE32+`, 204 KB) with the regenerated `resources.h`
- Extracted the embedded `RES_SCRIPT_JS_GZ` back out of `resources.h`, gunzipped: **byte-identical** to the parts concat, and normalized-identical to the pristine monolith
- `pnpm upload:local -- --mode=prod` — full snapshot OK; `pnpm test:hash` — JS↔C parity + snapshot provenance green
- `pnpm lint` / `pnpm format` / `pnpm test` (504/501/0, 3 by-design skips)

Closes out the §3.1 modularity series — no items remain on that audit row.


onemen:
🤖 AI review triage (Codebuff agent — local ADR 0020 review + CodeRabbit review:batch):

**review:local** — the provider free tiers were heavily rate-limited tonight; eight attempts each reviewed a partial 2-file subset (interleaved: embed.mjs, Makefile, DEVELOPING.md, .prettierignore — never the same subset twice) and returned **0 findings every time**. Disclosed as partial coverage rather than claimed as full.

**review:batch** — 1 major on `20-banners.js`: `ingestSelfUpdateSource` returns `true` even when the POST response is missing/`!res.ok`, so a server rejection stops the `ingestSelfUpdateSources` fallback chain to `pkg.selfUpdateUrl`. **Verified real but pre-existing**: the block is byte-carried from the pristine monolith (main's `script.js:498-501`) — my split moved it verbatim. Fixed here anyway rather than deferred, since the split is what surfaced it: the handler now returns `res && res.ok` so the fallback URL is tried on rejection, with `selfUpdateIngested` still set only on success.

Fixed in the follow-up commit; the served-asset byte-identity claim in the PR body applies to `00/10/30/40/50` — `20-banners.js` carries this one behavior fix.

PR #239, opened 2026-09-17
feat(release): partial publishes + a pinned, provenance-proven Windows toolchain

### LINKED ISSUE #157: AV false positive on installer_win.exe — tracking (WDSI case + SignPath signing)
# AV false positive on installer_win.exe — tracking (WDSI case + SignPath signing)

Tracking issue for the Windows Defender false positive on `installer_win.exe` and the
code-signing path that makes it stop recurring. Companion local draft:
`signpath.publish-win-wiring.local.yml` in the main checkout (gitignored; full content
embedded at the bottom of this issue).

## Timeline

- **2026-09-05** — prod publish of `installer_win.exe` (sha256 `76109ec3…`). Microsoft Defender flags it
  (`Program:Script/Wacapew.A!ml`, later surfaced as `Trojan:Win32/Wacatac.C!ml` on VirusTotal — a
  **machine-learning heuristic**). Multiple engines show the classic unsigned-binary ML-FP profile
  (McAfee local/cloud discrepancy, Bkav consistent; 60+ engines clean).
- **Root cause (confirmed)** — a **toolchain bump, not a code change**: the publish workflow ran
  `msys2/setup-msys2` with `update: true`, and a full `pacman -Syu` on 2026-09-05 pulled gcc
  **16.1.0 → 16.2.0** (published the day before). The recompiled bytes landed in the ML model's
  detection pocket. Every local rebuild (previous / exact-era / current source) scans clean.
- **2026-09-05/06** — `latest` release assets **deleted** so users stop testing flagged binaries
  (publish is now blocked until resolved by the new gate).
- **2026-09-06/07** — PR #151 lands the hardening stack (see below); **merged to main 2026-09-07 12:16 UTC**. Dev-mode publish on
  `av-false-positive-fixes` confirmed: `helper_win.exe` clean (0/60), `installer_win.exe` vetoed by
  Microsoft + Bkav → gate exits 1, **zero uploads** on fail.

## Fixes (PR #151 — merged 2026-09-07)

- [x] Merge PR #151 — PE metadata (`installer.rc`/`.manifest`, version resource), MSYS2 toolchain
      pin (`update: false`), host-AV gate (`tools/scan-av.mjs`), VirusTotal gate
      (`tools/scan-vt.mjs`, poll fix + **Microsoft veto** → abort publish on flag), two-pass
      SignPath signing skeleton in `upload.mjs` (`--build-only` / `--skip-build` +
      `dist/.build/build-manifest.json`), README/docs/license/CI coverage.

## Microsoft WDSI submission (short-term track)

| | |
|---|---|
| File | `installer_win-dev.exe` (CI-built dev binary, sha256 `bb4a6126…`) |
| Submission ID | `81e43d7f-1d0c-4798-88c2-ba05574f9034` |
| Status | Submitted — **Incorrect detection** |
| Submitted by | tabmix.onemen@gmail.com |
| Submitted | Sep 7, 2026 12:21 PM |
| Tracking URL | https://microsoft.com/en-us/wdsi/submission/81e43d7f-1d0c-4798-88c2-ba05574f9034 |
| Local record | `microsoft.submmision.local.log` (main checkout, gitignored) |

- [x] WDSI submission filed (Medium priority)
- [ ] WDSI closes as corrected → re-check hash on VirusTotal: want **Microsoft → undetected**
- [ ] Only then re-publish `latest` (the gate vetoes while Microsoft still flags)

## SignPath Foundation (durable track)

- [x] Application submitted — Sep 7, 2026 (confirmation received; review "within a few business days")
- [ ] Approval → SignPath dashboard: project `firefox-scripts`, artifact config (direct PE),
      signing policy `release-signing` on the shared OSS cert, GitHub connector/App
- [ ] Add secrets `SIGPATH_ORG_ID` + `SIGPATH_API_TOKEN` (org settings) + `.env-example` entries
- [ ] Wire `publish-win` signing per the embedded draft (two-pass: `--build-only` → upload →
      SignPath submit → copy signed back → `--skip-build` gates+publish on the signed bytes)
- [ ] Validate: dev publish → `signtool verify` → `pnpm scan:vt` on signed bytes → prod publish
- [ ] Download-page mention of SignPath Foundation — **done** (README + `latest` release body)

## Signing context (why this fixes the class)

Unsigned MinGW-gcc PEs embedding resources + a loopback HTTP server match the `!ml` structural
profile. A valid Authenticode signature (SignPath Foundation OSS cert) removes the heuristic
surface, is transparent to the updater (hashes track source trees, not binary bytes), and the
AV/VT gates run on the **signed** bytes in pass 2 — so the Microsoft veto keeps guarding exactly
what ships.

<details>
<summary>Embedded local draft: <code>signpath.publish-win-wiring.local.yml</code> (full content)</summary>

```yaml
# =============================================================================
# SignPath code signing — publish-win wiring draft  (LOCAL DRAFT, gitignored)
# =============================================================================
# Status: SignPath Foundation application submitted ~Sep 7 2026 (awaiting
#   approval). WDSI case 81e43d7f-1d0c-4798-88c2-ba05574f9034 (installer_win-dev.exe,
#   "Incorrect detection") is the short-term track.
#
# This file is the concrete plan for wiring SignPath signing into the
# publish-win job of .github/workflows/pages.yml ONCE the Foundation approves
# and the org secrets exist. Nothing here is wired yet — do not act on it until
# the prerequisites are true.
#
# The upload.mjs half of this plan already exists on PR #151 (commit 56554a3,
# tools/publish/upload.mjs): the --build-only / --skip-build flags + the
# dist/.build/build-manifest.json writer, unit-tested and smoke-tested two-pass.
# What this file drafts is the pages.yml workflow restructure that consumes it.
# PR #151 must be merged to main before this wiring can land.
# =============================================================================

# -----------------------------------------------------------------------------
# Prerequisites (SignPath side, after the Foundation approves)
# -----------------------------------------------------------------------------
# 1. SignPath.io dashboard: create project "firefox-scripts", an artifact
#    configuration for a direct PE (not a zip), and a signing policy
#    "release-signing" bound to the Foundation's shared OSS certificate.
# 2. Add the GitHub connector + install the SignPath GitHub App on
#    onemen/firefox-scripts (their source/build policy checks need it).
# 3. Create an API token with submitter rights scoped to project + policy.
# 4. Record the values:
#      SIGPATH_ORG_ID    -> Actions secret (like VT_API_KEY today)
#      SIGPATH_API_TOKEN -> Actions secret
#      project-slug      -> firefox-scripts   (NOT secret, committed in YAML)
#      signing-policy    -> release-signing   (NOT secret, committed in YAML)
#    .env-example should document the two local keys too, matching how
#    VT_API_KEY is handled.

# -----------------------------------------------------------------------------
# Design notes (why the flow looks like this)
# -----------------------------------------------------------------------------
# * SignPath's GitHub integration signs *workflow artifacts*: the exe must
#   first be uploaded with actions/upload-artifact, then submitted via
#   signpath/github-action-submit-signing-request (by artifact-id), then the
#   signed file is extracted from the action's output directory.
# * So publish-win becomes two passes around a signing step:
#     pass 1:  node upload.mjs --build-only   (build + stage, write manifest,
#              exit before the AV/VT gates + publish; staging kept on disk)
#     sign:    upload staged exes as artifacts -> SignPath -> copy signed
#              bytes back over the SAME staged paths from the manifest
#     pass 2:  node upload.mjs --skip-build    (AV + VT gates + publish run
#              on the exact signed bytes)
# * The AV/VT gates scan the staged file paths, so pass 2 automatically gates
#   the SIGNED bytes — the Microsoft veto keeps guarding exactly what ships.
# * Signing is transparent to the updater: hashes.json tracks source trees,
#   not binary bytes, so signed binaries do not disturb hash/update logic.
# * When SIGPATH secrets are absent the job degrades to today's single pass
#   (unsigned, still AV/VT-gated). The Microsoft veto would block a prod
#   publish of an unsigned installer until WDSI clears — intended.
# * Quota: as drafted, signing runs whenever secrets exist (prod AND dev).
#   If the Foundation's quota is tight, add `inputs.mode == 'prod'` to the
#   signing-step conditions (dev builds stay unsigned but still gated).
# * Both win binaries are signed (installer_win.exe + helper_win.exe). The
#   helper was never flagged, but it is equally unsigned and is dropped by the
#   updater into profile dirs — signing it is the durable fix for both.
# * The win job currently rebuilds in place on a windows-latest runner; the
#   signing steps below run on the same runner, so no cross-OS artifact
#   shipping is needed beyond the per-file upload/submit round trip.

# -----------------------------------------------------------------------------
# Replacement publish-win job (paste under jobs:, replacing the current one)
# -----------------------------------------------------------------------------
# Current state to preserve: the baseline-hashes download, the MSYS2 setup
# with `update: false`, and the shared ARGS composition (mode/platform/force +
# FIREFOX_SCRIPTS_STORED_HASHES_FILE for prod).
publish-win:
  name: publish — windows binaries (SignPath-signed when configured)
  needs: baseline
  environment:
    name: ${{ inputs.mode == 'prod' && 'release' || 'dev' }}
  runs-on: windows-latest
  env:
    # Job-level env so steps can branch on presence. Empty string when the
    # secret is not configured => job runs exactly like today (no signing).
    SIGPATH_API_TOKEN: ${{ secrets.SIGPATH_API_TOKEN }}
    SIGPATH_ORG_ID: ${{ secrets.SIGPATH_ORG_ID }}
  steps:
    - uses: $/.github/actions/setup-repo
      with:
        install: 'false'

    - name: Set up MSYS2 toolchain (make + gcc)
      # update: false — see the Wacapew false-positive note in the current job.
      uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2
      with:
        msystem: UCRT64
        update: false
        install: >-
          mingw-w64-ucrt-x86_64-gcc make

    - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
      with:
        name: baseline-hashes
        path: ${{ runner.temp }}/baseline/

    - run: pnpm install --frozen-lockfile

    # ---- shared upload args (compose once) ----
    - shell: bash
      run: |
        ARGS="--mode=${{ inputs.mode }} --platform=win"
        if [ "${{ inputs.force }}" = "true" ]; then ARGS="$ARGS --force"; fi
        if [ "${{ inputs.mode }}" = "prod" ]; then
          echo "FIREFOX_SCRIPTS_STORED_HASHES_FILE=$RUNNER_TEMP/baseline/hashes.json" >> "$GITHUB_ENV"
        fi
        echo "ARGS=$ARGS" >> "$GITHUB_ENV"

    # ---- PASS 1: build + stage, stop before gates/publish (only when signing) ----
    # Writes dist/.build/build-manifest.json and KEEPS the staging tree.
    - name: Build & stage binaries (pass 1)
      if: env.SIGPATH_API_TOKEN != '' && env.SIGPATH_ORG_ID != ''
      shell: bash
      env:
        VT_API_KEY: ${{ secrets.VT_API_KEY }}   # irrelevant in pass 1; harmless
      run: node tools/publish/upload.mjs $ARGS --build-only

    # ---- Read staged exe paths from the manifest for the signing steps ----
    # Manifest shape (per staged file): {role, platform, asset, absPath}
    #   installer_win.exe / helper_win.exe for platform=win.
    - name: Resolve staged exes from build manifest
      id: staged
      if: env.SIGPATH_API_TOKEN != '' && env.SIGPATH_ORG_ID != ''
      shell: bash
      run: |
        if [ ! -f dist/.build/build-manifest.json ]; then
          echo "installer=" >> "$GITHUB_OUTPUT"
          echo "helper=" >> "$GITHUB_OUTPUT"
          exit 0
        fi
        node -e '
          const m = require("./dist/.build/build-manifest.json");
          const f = p => (m.files.find(x => x.platform === "win" && x.role === p) || {}).absPath || "";
          console.log("installer=" + f("installer"));
          console.log("helper=" + f("helper"));
        ' >> "$GITHUB_OUTPUT"

    # ---- SIGN: per exe -> GH artifact -> SignPath request -> copy signed back ----
    - name: Upload installer artifact for signing
      id: up-installer
      if: steps.staged.outputs.installer != ''
      uses: actions/upload-artifact@<PIN-SHA> # vX — pin at wiring time; must expose artifact-id output
      with:
        name: unsigned-installer_win
        path: ${{ steps.staged.outputs.installer }}
        if-no-files-found: error

    - name: Submit installer signing request
      if: steps.staged.outputs.installer != ''
      uses: signpath/github-action-submit-signing-request@<PIN-SHA> # current release; pin at wiring time
      with:
        api-token: ${{ env.SIGPATH_API_TOKEN }}
        organization-id: ${{ env.SIGPATH_ORG_ID }}
        project-slug: firefox-scripts
        signing-policy-slug: release-signing
        github-artifact-id: ${{ steps.up-installer.outputs.artifact-id }}
        wait-for-completion: true
        output-artifact-directory: ${{ runner.temp }}/signed-installer

    - name: Copy signed installer over staged path
      if: steps.staged.outputs.installer != ''
      shell: bash
      run: |
        cp "$RUNNER_TEMP/signed-installer/installer_win.exe" \
           "${{ steps.staged.outputs.installer }}"

    - name: Upload helper artifact for signing
      id: up-helper
      if: steps.staged.outputs.helper != ''
      uses: actions/upload-artifact@<PIN-SHA> # vX — pin at wiring time; must expose artifact-id output
      with:
        name: unsigned-helper_win
        path: ${{ steps.staged.outputs.helper }}
        if-no-files-found: error

    - name: Submit helper signing request
      if: steps.staged.outputs.helper != ''
      uses: signpath/github-action-submit-signing-request@<PIN-SHA>
      with:
        api-token: ${{ env.SIGPATH_API_TOKEN }}
        organization-id: ${{ env.SIGPATH_ORG_ID }}
        project-slug: firefox-scripts
        signing-policy-slug: release-signing
        github-artifact-id: ${{ steps.up-helper.outputs.artifact-id }}
        wait-for-completion: true
        output-artifact-directory: ${{ runner.temp }}/signed-helper

    - name: Copy signed helper over staged path
      if: steps.staged.outputs.helper != ''
      shell: bash
      run: |
        cp "$RUNNER_TEMP/signed-helper/helper_win.exe" \
           "${{ steps.staged.outputs.helper }}"

    # ---- PASS 2: AV + VT gates + publish on the SIGNED bytes ----
    # No signing configured => this single pass also does the build (ARGS has
    # no --build-only/--skip-build), i.e. exactly today's behavior.
    - name: AV/VT gates + publish (pass 2, or single pass when not signing)
      shell: bash
      env:
        VT_API_KEY: ${{ secrets.VT_API_KEY }}
      run: |
        if [ -n "$SIGPATH_API_TOKEN" ] && [ -n "$SIGPATH_ORG_ID" ]; then
          node tools/publish/upload.mjs $ARGS --skip-build
        else
          node tools/publish/upload.mjs $ARGS
        fi

# publish-linux and publish-mac are unchanged (still needs: publish-win).

# -----------------------------------------------------------------------------
# After wiring — validation sequence (do in this order)
# -----------------------------------------------------------------------------
# 1. Dev publish from a branch (`gh workflow run pages.yml -f mode=dev`):
#    watch build -> sign -> AV -> VT on the signed bytes.
# 2. Confirm the signature: `signtool verify /pa installer_win.exe` on the
#    signed exe (OSS "test-signing" policy first if the Foundation offers it,
#    then release-signing for prod). Verify the chain resolves on a clean
#    Windows box (SignPath Foundation OSS CA must be trusted — open question
#    to confirm on first signed build).
# 3. Re-check the signed bytes: `pnpm scan:vt` locally + the gate output.
#    Expect Microsoft/Bkav to go quiet (signed => ML heuristic no longer
#    matches the unsigned-binary profile).
# 4. Only then prod-publish `latest`. Keep the WDSI case open in parallel —
#    it covers users who already downloaded the flagged unsigned hash.
# 5. Watchdog/CI: the pages.yml change needs docs/ci-inventory.md sync if any
#    workflow name/trigger/gate semantics change.

# -----------------------------------------------------------------------------
# Open questions to resolve at wiring time
# -----------------------------------------------------------------------------
# * Which signpath/github-action-submit-signing-request version to pin (v1 in
#   SignPath's own demo; check for a newer one that accepts artifact-path
#   directly, which would let us drop the upload-artifact round trip).
# * upload-artifact version must expose the artifact-id output (v4+ does).
# * Whether signing should run for dev mode too or be prod-gated (quota).
# * The SignPath OSS certificate is issued in the Foundation's name — the
#   "Download page must mention the SignPath Foundation" term is already met
#   (README + latest release body, added on PR #151).
```
</details>

onemen:
Fixes #157 (release-blocking AV false positive) by making the release *splittable*, and makes the
Windows binaries reproducible by pinning the toolchain that builds them.

## 1. Partial publishes — `--skip=<role>` (ADR [0030](./docs/decisions/0030-partial-publishes.md))

A flag usually hits one binary, not the packages. A held-back role is not built, hashed, scanned or
uploaded, and its `hashes.json` entry stays **frozen** at the last published value — so nothing
installed is ever pointed at bytes that were never published. Roles are independent, and the AV/VT
gates still cover exactly what ships.

```bash
pnpm upload:local -- --mode=prod --skip=installer,helper   # zips + hashes.json only
pnpm upload -- --mode=dev --skip=installer                 # clean helper, withheld installer
gh workflow run pages.yml -f mode=prod -f skip=installer,helper
```

Proven end-to-end locally: with a crafted baseline (`installer=SENTINEL…, helper=SENTINEL…`),
`--skip=helper` recomputed the installer entry while the helper stayed `SENTINEL-HELPER`;
`--skip=installer,helper` froze both and logged *"no binaries in scope — the AV/VT gates had nothing
to scan"*.

## 2. Pinned MSYS2 toolchain + provenance (`config/msys2-toolchain.json`, `pnpm toolchain:local`)

The published PEs are unsigned with a per-hash AV verdict, so the bytes CI ships must be the bytes a
developer can rebuild. They were not, and a runner dispatch shows precisely why:

```
✓ 5 tools from d:/a/_temp/msys64/ucrt64/bin report the pinned versions
  ✓ gcc     D:/a/_temp/msys64/ucrt64/bin/gcc.exe sha256:f96a3bdb1d3a   (16.1.0)
      also on PATH: C:/mingw64/bin/gcc.exe, C:/Strawberry/c/bin/gcc.exe
```

`msys2/setup-msys2` defaults to `path-type: minimal` and extracts MSYS2 outside `C:\msys64`, so the
build's `make` was MSYS2's while `gcc`/`ld`/`windres` came from **`C:/mingw64/bin`** — the runner
image's mingw. Rebuilding the flagged commit with the toolchain its own log recorded differs by
2,752 bytes of `.text` (docs/DEVELOPING.md → "Why a clean local scan does not clear a CI build").

The action now installs all 18 pinned packages with `pacman -U`, asks the msys2 shell where it
actually lives (`cygpath -m /` → `MSYS2_LOCATION`) to put that tree first on the build steps' PATH,
and then proves it: `--provenance` fails the job unless `gcc`/`ld`/`as`/`windres`/`make` all report
the pinned versions from one directory. `pnpm toolchain:local` extracts the same files into
`dist/.toolchain/` for a byte-reproducible local build, and the `deterministic` job builds one commit
with the environment toolchain *and* with that prefix, requiring byte-identical output.

Two byte inputs beyond the MSYS2 pin are handled explicitly: the `node`/`zlib` pair that gzips the
embedded web assets (logged by `--provenance`, pinned by a unit test with an actionable failure
message) and the `dev-build-<branch>-<sha>` config string, which a detached checkout renders as
`HEAD` (documented, with `FIREFOX_SCRIPTS_REF_NAME` to match CI).

## 3. What the first runner dispatches found (all fixed here)

| Red run | Cause | Fix |
| --- | --- | --- |
| `node: command not found` (127) | `shell: msys2 {0}` drops the toolcache from PATH | the pin runs in `shell: bash`, where node exists |
| `pacman -U failed (exit null)` | MSYS2 is not at `C:\msys64`, so pacman fell back to a PATH that lacks it | the msys2 shell publishes its own root; pacman is called by absolute path |
| `tar: Cannot connect to D: resolve failed` | MSYS2's GNU tar reads `D:/…` as an rsh host | archive named by basename with the cache dir as cwd; `-C` forward-slashed |
| local `--prefix` extracted into a dir called `true` | bare `--prefix` was read as its own value | `optionValue()` distinguishes bare/valued/absent |

A missing install now fails with the candidates it tried and the env vars it saw, instead of looking
like a pacman crash.

## Evidence from the runner

* **Provenance passes**, and the pinned binaries build the artifacts: the dev publish reached the
  VirusTotal gate with `installer_win-dev.exe` built by gcc 16.1.0 / binutils 2.46.
* **The same-runner determinism pair passes**: *"snapshots byte-identical"*.
* **VirusTotal vetoes the installer anyway** — 2/70 engines (Elastic, Microsoft), and Microsoft is a
  veto engine, so the publish correctly refuses; `helper_win-dev.exe` is clean 0/70. This is a
  **content** verdict on freshly pinned bytes, not a toolchain artifact: pinning buys
  reproducibility, so WDSI submission + signing (SignPath) remain the levers, and `--skip=installer`
  is the way to keep shipping the parts that are clean. The host Defender gate produced no signal on
  the runner (it reported both files as *skipped*).

## Validation

`pnpm lint`, `pnpm format`, `pnpm test` (549 tests, 546 pass / 3 pre-existing skips),
`pnpm check:decisions`. Locally verified: `pnpm toolchain:local` extracts the pinned gcc 16.1.0 /
binutils 2.46 and builds prod binaries with them; `pnpm upload:local` frozen-entry semantics for
`--skip`. Not verified: `pnpm test:hash` (hash path untouched), and the new env-vs-prefix byte comparison
passes on the runner — installer_win.exe and helper_win.exe are byte-identical between the
environment build and the pinned prefix (run 35274369030).

🤖 Generated with Codebuff


onemen (commented):
🤖 AI review triage (Codebuff agent) — findings record for this branch

Every finding gathered while building, dispatching and debugging this branch, so the evidence lives
next to the code instead of in a chat log. Line-anchored review threads come from the separate
`pnpm review:local` pass (ADR 0020).

## 1. Pinned toolchain verified on the runner

`pages.yml` run [35272125354](https://github.com/onemen/firefox-scripts/actions/runs/35272125354),
`publish — windows binaries`, step *Prove the build shell resolves the pinned toolchain*:

```
pinned MSYS2 toolchain — 18 package(s), captured from gh run 34935654811
  pinned: gcc 16.1.0, binutils 2.46, make 4.4.1
  runtime: node 24.20.0, zlib 1.3.2.1-motley-42c2f19 (gzip inputs of resources.h)
  ✓ gcc     D:/a/_temp/msys64/ucrt64/bin/gcc.exe sha256:f96a3bdb1d3a
      gcc (Rev5, Built by MSYS2 project) 16.1.0
      also on PATH: C:/mingw64/bin/gcc.exe, C:/Strawberry/c/bin/gcc.exe
  ✓ ld      D:/a/_temp/msys64/ucrt64/bin/ld.exe sha256:a05da6de92dc      (GNU ld 2.46)
  ✓ as      D:/a/_temp/msys64/ucrt64/bin/as.exe sha256:c7ca4e94db50      (GNU assembler 2.46)
  ✓ windres D:/a/_temp/msys64/ucrt64/bin/windres.exe sha256:ec1898ea1776  (GNU windres 2.46)
  ✓ make    D:/a/_temp/msys64/usr/bin/make.exe sha256:91b7e155590d       (GNU Make 4.4.1)
✓ 5 tools from d:/a/_temp/msys64/ucrt64/bin report the pinned versions
```

## 2. Byte reproducibility is now CI-asserted

`build-and-upload.yml` run [35274369030](https://github.com/onemen/firefox-scripts/actions/runs/35274369030),
`deterministic publish output` — **success**, both halves:

```
snapshots byte-identical
installer_win.exe: pinned prefix build == environment build
helper_win.exe:    pinned prefix build == environment build
```

The second pair is the claim this branch exists for: the same commit built with the pacman-installed
toolchain and with the tables extracted by `pnpm toolchain:local` is byte-for-byte the same, so a
local rebuild of a published commit reproduces the published hashes.

## 3. Root cause of the #157 irreproducibility (proven, not inferred)

* Run `34935654811` **installed exactly the 18 packages this branch pins** — its log lists
  `mingw-w64-ucrt-x86_64-binutils-2.46-4 … make-4.4.1-3 mingw-w64-ucrt-x86_64-gcc-16.1.0-5`.
* Rebuilding its commit `e393191` with that toolchain does **not** reproduce the uploaded bytes:
  `installer_win-dev.exe` 202,752 B vs 199,168 B (`.text` 127,904 vs 125,152 = +2,752 B), and the
  helper links one extra CRT import, `api-ms-win-crt-multibyte-l1-1-0.dll`; at the same source ~94%
  of `.text` differs → a different compiler, not a flag.
* The provenance step names it: `C:/mingw64/bin/gcc.exe` (and `C:/Strawberry/c/bin`) sit on the
  runner PATH behind MSYS2. `msys2/setup-msys2` defaults to `path-type: minimal` and extracts to
  `D:/a/_temp/msys64`, so `ucrt64/bin` was never on the job PATH: `make` was MSYS2's while
  `gcc`/`ld`/`windres` were the image's mingw. Those published bytes are unattributable and their
  VT/WDSI verdicts cannot be re-derived; the pin's output is the reproducible reference.
* The pinned prefix is deterministic *and* hermetic: two consecutive builds are byte-identical and
  were produced with nothing but the extracted packages.
* Config-level gotcha, now documented: the generated `_config.h` bakes `dev-build-<branch>-<sha>`,
  and a detached checkout (`git worktree add`, `git checkout <sha>`) reports `HEAD` — so a dev-mode
  rebuild from a worktree carries `dev-build-HEAD-<sha>` URLs where CI's carries
  `dev-build-main-<sha>`. Export `FIREFOX_SCRIPTS_REF_NAME=<branch>` to match.

## 4. The second byte input: node/zlib

`installer/embed.mjs` gzips the embedded web assets with Node's bundled zlib, so `resources.h` — and
the installer's bytes and hashes — depends on the runtime the MSYS2 pin cannot cover. Measured
2026-09-17: node 24.20.0 (zlib `1.3.2.1-motley-42c2f19`, what the flagged run used) and node 26.8.2
(zlib `…-5eb4d7e`) emit **identical** gzip (index.html 2,801 B, style.css 5,498 B), so this is a
latent input rather than an active mismatch. `--provenance` logs the pair on every build, and
`test/unit/installer/embed.test.mjs` pins the compression behaviour with a failure message that names
the cause and the next step. The `deterministic` job cannot see this class of drift (its two builds
share one Node) — the log is what makes it visible.

## 5. AV findings from the dev-publish dispatch

Run [35272125354](https://github.com/onemen/firefox-scripts/actions/runs/35272125354):

* **Host Defender on the runner produced no signal**: it reported both binaries as *skipped*
  (`AV scan skipped for …: Scanning … was skipped`), so the only AV coverage was VirusTotal.
* **VT vetoes the installer**: `installer_win-dev.exe` — 2 malicious / 68 undetected, flagged by
  **Elastic and Microsoft**; Microsoft is a veto engine, so the run correctly refuses to publish.
  `helper_win-dev.exe` was clean, 0/70.
* These are **freshly built, pinned-toolchain bytes with a new hash**, so the Microsoft false
  positive is **content-driven, not toolchain-driven**. Pinning buys reproducibility, not a clean
  verdict: WDSI submission and code signing (SignPath) remain the levers, and `--skip=installer` is
  how the clean parts keep shipping in the meantime.
* Nothing was published by that run — the refusal happens before any push (the only `dev-build-*` ref
  on the remote is the pre-existing `dev-build-main-450468f`).

## 6. AV-shape work also in this branch

* The helper had a version resource but **no manifest**; it now ships `installer/src/helper/helper.manifest`
  (asInvoker + Win10/11 compatibility GUIDs) instead of relying on the crt's auto-linked
  `default-manifest.o`, which CI's crt did not supply. A unit test pins both `.rc` resources to
  `installer.conf`'s `BUILD_DATE`, and the metadata fields are populated.
* Import tables measured: installer → `ADVAPI32, bcrypt, KERNEL32, api-ms-win-crt-*, SHELL32, USER32,
  WS2_32`; `crypt32` is **not** imported (so `-lcrypt32` in the link line is a no-op, left as is).
  No packer is used (`-s` strip only).
* Remaining inherent heuristic surfaces, for whoever picks up the AV work: the loopback HTTP server,
  high-entropy embedded zips in `.rdata`, and `certutil -hashfile` spawned with redirected handles
  and temp files in `detect_browser.c` (a LOLBin pattern). PE subsystem is still 5.02 — the 6.0 bump
  (#160) was reverted by #161, so shape edits are a lottery, not a dial.

## 7. Four red runs on the way here (all fixed in this branch)

| Red run | Symptom | Cause | Fix |
| --- | --- | --- | --- |
| `35269902992` | `node: command not found`, exit 127 | `shell: msys2 {0}` drops the toolcache from PATH | pin runs in `shell: bash` |
| `35271067930` | `pacman -U failed (exit null)` | MSYS2 is not at `C:\msys64`; pacman fell back to a PATH without it | the msys2 shell publishes its own root (`cygpath -m /` → `MSYS2_LOCATION`); pacman by absolute path |
| `35273187044` | `tar: Cannot connect to D: resolve failed` | MSYS2's GNU tar reads `D:/…` as an rsh host (invisible to local Git-Bash's bsdtar) | archive by basename with the cache dir as cwd; `-C` forward-slashed |
| local run | `--prefix` extracted into a directory called `true` | bare `--prefix` was read as its own value | `optionValue()` distinguishes bare/valued/absent; valueless `--require-root` is now an error |

A missing install now fails naming the candidates it tried and the env vars it saw, rather than
looking like a pacman crash.

## 8. Validation status

Green: `pnpm lint`, `pnpm format`, `pnpm test` (549 tests, 546 pass / 3 pre-existing skips),
`pnpm check:decisions`, and on CI the whole `deterministic` job plus all three build legs. Verified
locally: `pnpm toolchain:local` extracts the pinned gcc 16.1.0 / binutils 2.46 and builds prod
binaries with them; `--skip` frozen-manifest semantics with a crafted baseline. Not verified:
`pnpm test:hash` (hash path untouched) and any real publish (each one is operator-initiated).

tools/ci/msys2Toolchain.mjs
```diff
@@ -0,0 +1,708 @@
...
+    }
+    if (pkg.role !== 'mingw' && pkg.role !== 'system') {
+      problems.push(`${label}: role must be 'mingw' or 'system'`);
+    }
+    // A repo URL that does not match the package's own prefix is the classic
+    // copy-paste slip in a pinned manifest (msys packages live under /msys).
+    if (
+      repos?.[pkg.repo] &&
+      pkg.arch &&
+      !packageFileName(pkg).endsWith(`-${pkg.arch}.pkg.tar.zst`)
```
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**Accepted — right (minor).** This check is dead code. `packageFileName(pkg)` *builds* the name as `` `${pkg.name}-${pkg.version}-${pkg.arch}.pkg.tar.zst` `` (line 72), so `packageFileName(pkg).endsWith(\`-${pkg.arch}.pkg.tar.zst\`)` is a tautology: the branch is unreachable and `arch '…' does not match the file name` can never be reported. The comment above it describes a check that does not exist — nothing here looks at the repo URL.

Three slips this fail-closed gate lets through:

- an **undeclared repo** — `repos?.[pkg.repo] &&` short-circuits the check; the mistake only surfaces later, when `packageUrl` throws mid-download;
- a **missing/blank `arch`** — the condition is skipped silently, and `packageFileName` would then produce `…-undefined.pkg.tar.zst`;
- a **`mingw-w64-*` package pointed at a non-`ucrt64` repo** — line 121 only checks the reverse direction (`repo === 'ucrt64'` ⇒ name prefix), which is the copy-paste slip the comment claims to catch.

Suggested fix, using the symbols already in scope: drop the tautology and assert, per package — `!repos?.[pkg.repo]` → `unknown repo '<repo>'`; `!pkg.arch` → `arch is required`; `pkg.repo !== 'ucrt64' && pkg.name.startsWith('mingw-w64-')` → `mingw-w64-* packages must use the ucrt64 repo`; then repair the comment. Add the matching `validateManifest` rejections beside the existing ones — the current tests cover the duplicate, sha256, version, role and `ucrt64` name-prefix cases, but nothing asserts this branch, which is exactly why a check that can never fire went unnoticed.

No fix is included in this review run; this thread is resolved when the fix lands.

.github/actions/pinned-msys2/action.yml
```diff
@@ -0,0 +1,88 @@
...
+      # `msys64\usr\bin\bash.EXE` instead of Git-Bash, since this shell's dirs
+      # precede the system PATH. The publish flow was verified end-to-end under
+      # it (build → zips → upload → VT gate); MSYS2's GNU tar is stricter than
+      # Git-Bash's bsdtar about drive-letter file arguments, which is why
+      # tools/ci/msys2Toolchain.mjs passes archive paths colon-free.
+      run: |
+        echo "MSYS2_LOCATION=$(cygpath -m /)" >> "$(cygpath -u "$GITHUB_ENV")"
+        echo "PINNED_TOOLCHAIN_PREFIX=$(cygpath -m "$MINGW_PREFIX/bin")" >> "$(cygpath -u "$GITHUB_ENV")"
+        cygpath -w "$MINGW_PREFIX/bin" >> "$(cygpath -u "$GITHUB_PATH")"
+        cygpath -w /usr/bin >> "$(cygpath -u "$GITHUB_PATH")"
```
onemen, line 78: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

**Accepted — right.** `GITHUB_PATH` is not first-line-wins: the runner collects every line into `Global.PrependPath` in file order (`AddPathFileCommand.ProcessCommand`, `src/Runner.Worker/FileCommandManager.cs`), then joins that list **reversed** into the string it puts in front of `PATH` — `string.Join(sep, ExecutionContext.Global.PrependPath.Reverse<string>())` at `src/Runner.Worker/Handlers/Handler.cs:216`, applied by `PathUtil.PrependPath`. So the **last** line written wins.

Consequence here: the write order is `$MINGW_PREFIX/bin` (line 68) then `/usr/bin` (line 69), which leaves the pinned UCRT64 directory *behind* `/usr/bin` on `PATH` — the opposite of this action's stated intent ("those dirs go first") and of what the step's own echo claims (`using the existing cygpath commands`). Swapping the two writes, as suggested, is what actually puts `$MINGW_PREFIX/bin` first.

Why it has not bitten yet: on the runner `/usr/bin` carries no `gcc`/`ld`/`as`/`windres` — the green run's `--provenance` output lists every other match on `PATH` for those tools (`C:/mingw64/bin/*`, `C:/Strawberry/c/bin/*`) and none from `/usr/bin` — so today's resolution is identical in either order, and `make` (the one tool that legitimately lives in `/usr/bin`, pinned there as a `system` package) resolves to `/usr/bin/make.exe` as before. MSYS `binutils` *does* install `ld.exe`, `as.exe` and `windres.exe` into `/usr/bin` (a local MSYS2 install has all three), so the moment anything pulls that package into the bootstrap, the current order would hand the build unpinned MSYS tools instead of the pinned ones; `--provenance --require-root` would then fail the job rather than ship wrong bytes — a loud failure, but a self-inflicted one that points at the toolchain rather than at this ordering.

While fixing, two things so the next edit cannot re-introduce it: reword the echo on line 70 to state the resulting precedence rather than the write order (e.g. `pinned toolchain first on PATH: $MINGW_PREFIX/bin, then /usr/bin`), and record in the header comment that `GITHUB_PATH` entries win in reverse write order.

No fix is included in this review run; this thread is resolved when the fix lands.

onemen (commented):

Source: `pnpm review:batch --pr 239 --agent` — one batched CodeRabbit pass over `main...av/release-split` at f9fe9ca (`review_completed`, all 23 changed files, 2 findings). Triage: **2 accepted, 0 rejected**, each posted as its own line-anchored thread below.

- `tools/ci/msys2Toolchain.mjs:117` — minor — `validateManifest`'s arch check is a tautology (dead validation).
- `.github/actions/pinned-msys2/action.yml:69` — major — `GITHUB_PATH` entries win in *reverse* write order, so `/usr/bin` currently precedes the pinned UCRT64 directory.

Coverage note: the local `pnpm review:local` pass contributed one file only — the configured default model (`gemini-3.6-flash`) is out of free-tier quota (`GenerateRequestsPerDayPerProjectPerModel`, 20/day), the reviewer ends the whole run on the first 429, and the fallback provider is rate-limited as well. The CodeRabbit batch pass above is therefore the whole review; its severity labels are its own.

tools/ci/msys2Toolchain.mjs
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

Fixed in edf5d94. `validateManifest` now rejects an undeclared repo (`unknown repo 'x' (declared: …)`), a missing `arch` (`arch is required`), and either direction of the name/repo mismatch (`mingw-w64-* packages belong to the ucrt64 repo`); the tautological condition and the comment describing a check that did not exist are gone.

Tests: `validateManifest rejects an undeclared repo, a missing arch and a mingw package outside ucrt64`, plus `validateManifest accepts an msys package that is not a mingw-w64-* one` as the guard against an over-broad rule, and the shipped-manifest exact-equality test now covers the new checks. `pnpm test` 552 (549 pass / 0 fail / 3 pre-existing skips), `pnpm lint`, `pnpm format` all green.

.github/actions/pinned-msys2/action.yml
onemen: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run)

Fixed in 1fe603d. The two `GITHUB_PATH` writes are swapped — `/usr/bin` first, `$MINGW_PREFIX/bin` second — so the pinned UCRT64 dir is now the first entry on `PATH`. The step's echo and the header comment state the resulting order and why it is reversed, and a unit test reads the action and asserts the write order so it cannot be flipped back silently.

For the record, this changes nothing in the current runner's resolution: its `/usr/bin` carries no `gcc`/`ld`/`as`/`windres` (the run's own `--provenance` output lists every other `PATH` match for those tools and none from `/usr/bin`), which is exactly why the pin passed while the ordering was wrong. On an MSYS2 where binutils is installed, `/usr/bin` holds `ld`, `as` and `windres`, and the old order would have let those shadow the pinned ones.

docs/decisions/0030-partial-publishes.md
```diff
@@ -0,0 +1,46 @@
...
+veto) correctly refuses to ship a flagged binary, but until now that refusal froze **every**
+artifact in the run: the zips carry the script updates installed browsers consume (and their
+`hashes.json` entries), so one flagged `installer_win.exe` also stopped script delivery. The
+alternatives were worse — publish the flagged binary and hand users a Defender popup, or publish
+nothing at all.
+
+## Decision
+
+`tools/publish/upload.mjs` accepts `--skip=packages|installer|helper` (repeatable or
+comma-separated), and the CI dispatches (`pages.yml`, `build-and-upload.yml`) take the same list in
```
onemen: 🤖 AI review — agent-posted (ADR 0020) — high — **the documented CI `skip` input does not exist.**

ADR 0030 and docs/DEVELOPING.md instruct `gh workflow run pages.yml -f mode=prod -f skip=installer,helper`, and this Decision asserts "the CI dispatches (`pages.yml`, `build-and-upload.yml`) take the same list in a `skip` input" — but neither workflow defines a `skip` input on this branch (pages.yml inputs: `mode`, `force` only; all four `upload.mjs` steps build `$ARGS` from those alone), and `.agents/skills/publishing/SKILL.md` repeats the command.

Consequences: the dispatch fails with `unexpected input: skip` — or, if the operator dispatches without it, a prod run publishes **full**, shipping exactly the binary they believed withheld (the failure the flag exists to prevent).

Fix: add `skip: {description: 'roles to hold back (--skip): packages|installer|helper, comma-separated', default: ''}` to both `workflow_dispatch` blocks and thread it into every `$ARGS` (`SKIP: ${{ inputs.skip }}` in step env, then `if [ -n "$SKIP" ]; then ARGS="$ARGS --skip=$SKIP"; fi`). Mention the input in docs/ci-inventory.md's dispatch rows so it stays synchronized.

test/unit/installer/embed.test.mjs
```diff
@@ -62,8 +64,39 @@ test('embed.mjs: gzip byte arrays are non-empty and well-formed', () => {
...
+  const raw = Buffer.from(
+    Array.from(
+      {length: 64},
+      (_, i) => `line-${i}: the quick brown fox jumps over the lazy dog`
+    ).join('\n'),
+    'utf-8'
+  );
+  const gz = zlib.gzipSync(raw, {level: 9});
+  assert.equal(gz.length, 236, 'gzip output length changed for the pinned sample');
+  assert.equal(
```
onemen: 🤖 AI review — agent-posted (ADR 0020) — high — **the pinned gzip hash is runtime-dependent; CI already fails on this PR's own head.**

`lint + format` in run 35278240754: node v24.20.0 / zlib 1.3.2.1-motley-42c2f19 hashes this sample to `3c4e7be57283a1a2…` (length 236 matches; bytes differ), so the assertion fails and the CI gate is red. The constant was computed on the author's runtime — the premise "measured identical on 24.20.0 and 26.8.2" does not generalize, and this run is the falsifying datapoint.

Suggested fix: keep only runtime-independent invariants here (determinism: two `gzipSync` calls byte-equal; round-trip: `gunzipSync(gz)` equals `raw`), and leave cross-runtime equality to the `--provenance` node/zlib logging this PR adds plus a documented one-time compare. If a hash pin is wanted, key it by `process.versions.zlib` with an "unknown zlib — verify and add" failure instead of one absolute constant.

docs/decisions/0030-partial-publishes.md
```diff
@@ -0,0 +1,46 @@
...
+roles, and the AV/VT gates report that they had nothing in scope — every gate still covers exactly
+the bytes that ship. The no-`--skip` path is unchanged, and the roles are independent: withholding
+the installer can still ship a clean helper, which is what the in-browser updater needs for elevated
+copies to admin-protected install dirs.
+
+## Consequences
+
+Script delivery survives an AV holdback, and the withheld binary keeps serving its last published
+(accepted) bytes instead of regressing to nothing. A held-back role also skips its rebuild, so
+whenever its sources really did change the frozen entry stays stale by exactly one revision — which
```
onemen: 🤖 AI review — agent-posted (ADR 0020) — low — **"stale by exactly one revision" holds only when the next run is a full publish.**

Consecutive partial publishes that hold back the same role across multiple revisions accumulate staleness; the absolute phrasing can read as "repeated skips are safe". Suggested wording: the entry stays stale until the next full publish of that role (one revision when the very next run is full, longer under repeated holdbacks). The self-healing mechanism itself is sound.

tools/publish/upload.mjs
```diff
@@ -1023,6 +1087,12 @@ async function main() {
     // scanner only warns; a positive detection hard-fails the run.
     section('AV scan');
     const avFiles = [...builtInstallers.map(installerPath), ...builtHelpers.map(helperPath)];
+    if (avFiles.length === 0 && noBinaryScope(SCOPE)) {
```
onemen, line 1140: 🤖 AI review — agent-posted (ADR 0020) — medium — **`--skip=packages` on a dev publish strands the branch manifest.**

Dev path: `for (const name of builtZips) pagesFiles[zipFileName(name)] = …` (publishToGitHub) — with `--skip=packages` `builtZips` is empty, so the pushed `dev-build-*` branch gets the new `hashes.json` (utils/fx-folder/updater-ui entries frozen) **without** the zips those entries describe. A browser on that branch reports Update available and the installer 404s `utils.zip` — a state that can never converge, the exact stranding ADR 0030 says the freeze prevents. Local snapshots of such a run show the same shape (entries present, files absent).

Options: reject the combination (`--skip=packages` only in prod, where the existing release/branch keep serving the old bytes), or copy the previous zips into the dev branch when packages are held back. Whichever is chosen deserves a pinning test in publishScope.test.mjs.

docs/decisions/0030-partial-publishes.md
onemen: 🤖 AI review triage (Codebuff agent — follow-up on my AI review thread) — fixed in 7547698: `skip` is now a `workflow_dispatch` input on both pages.yml and build-and-upload.yml, threaded to every upload.mjs step through an `UPLOAD_SKIP` env var that appends `--skip=$UPLOAD_SKIP` when non-empty (dispatch inputs reach the shell only via env vars, per the workflow's own injection-surface rule). Resolving.

test/unit/installer/embed.test.mjs
onemen: 🤖 AI review triage (Codebuff agent — follow-up on my AI review thread) — fixed in 85385cc: the absolute sha256/length assertions are replaced with runtime-independent invariants (in-runtime determinism, lossless round-trip, gzip magic, plausible size band), and docs/DEVELOPING.md now documents the node/zlib pair as version-sensitive — reproduce published bytes from the pair `--provenance` records. Resolving.

tools/publish/upload.mjs
onemen: 🤖 AI review — agent-posted (ADR 0020) — fixed in 8e71874.

Verified the actual semantics against `uploadToPages.mjs`: `uploadFilesToPages` never deletes — it only adds/updates blobs on a preserved tree. So on an **existing** dev branch, a `--skip=packages` run keeps the prior zips serving and the frozen manifest stays truthful (consistent). The only strand-shaped case is a `--skip=packages` run creating a **brand-new** dev branch.

Implemented the fix as a loud warning instead of a rejection: `publishScope.devStrandingWarn()` probes the target branch via the GitHub API (probe failures downgrade to the generic note), and `upload.mjs` prints it at the scope banner — before any build — telling the operator the new branch would carry a manifest whose package entries 404 and to use a full run instead. Covered by tests in `publishScope.test.mjs`.

docs/decisions/0030-partial-publishes.md
onemen: 🤖 AI review — agent-posted (ADR 0020) — fixed in 8e71874.

ADR 0030 (and the matching DEVELOPING.md sentence) no longer says "stale by exactly one revision": it now says entries stay frozen **until the next full publish of that role**, with a caveat sentence noting the one stranding shape — a partial run creating a brand-new dev branch — which is now warned loudly at dispatch time.

PR #244, opened 2026-09-18
feat(installer): make verify checks PE structure; document the Defender write-lock race (#233)

### LINKED ISSUE #233: Dev environment: Defender real-time scan intermittently breaks local Windows links (collect2: ld returned 5, truncated exe)

onemen:
Closes #233 (remainder — the publish-side magic-byte check already landed in #234).

## What

- **`make verify` now validates PE structure before the GUI-subsystem check**: MZ header, `e_lfanew` in range, `PE\0\0` at the offset — failing loudly with a re-run hint when Defender's real-time scan holds a write lock on the freshly linked exe and truncates it (the #233 local-build failure mode).
- **`docs/DEVELOPING.md`** gains the workaround recipe: what the failure looks like (`collect2: ld returned 5`, truncated exe), retry guidance, and the real-time-scan exclusion for `dist/`; notes CI is unaffected (runners run Defender passive; the publish gate re-checks every staged binary).

## Validation

- Fresh `mingw32-make dist_win` + `mingw32-make verify` on the worktree: PE check passes (204,800 bytes), GUI subsystem confirmed.
- Failure path proven: truncating the built exe to 0 bytes makes `make verify` exit 2 with `✗ TRUNCATED/INVALID PE: no MZ header`.
- Pre-push gates green (eslint, tsc, format-c, gcc -fanalyzer, markdownlint, 561 unit tests).

Fixes #233


docs/DEVELOPING.md
```diff
@@ -105,6 +105,20 @@ uses a compatible `mkdir` either way, so the same targets work from cmd.exe, Pow
...
+
+**Defender write-lock race on local links (issue #233).** Windows Defender's real-time scan can
+intermittently hold a write lock on the freshly linked `installer_win.exe` while `collect2`/`ld` is
+still writing it — the link dies with `collect2.exe: error: ld returned 5 exit status` and/or the
+exe is left truncated (a few hundred bytes, no PE structure). `make verify` catches the truncated
+output immediately (MZ/e_lfanew/PE-signature check). If it fires:
+
+1. Just re-run the link — the race is transient and a retry usually succeeds.
+2. If it recurs, exclude the build output dir from real-time scanning (Windows Security → Virus &
+   threat protection → Exclusions → `<repo>\dist`, your repository's `dist` directory), or pause
```
onemen, line 118: 🤖 AI review triage (Codebuff agent — result of the review:batch run): accepted. The example hardcoded the author's checkout path. Fixed in 2a7a35f: the exclusion is documented as `<repo>\dist` (checkout-agnostic).

PR #245, opened 2026-09-18
fix(core): sweep stale chrome.manifest* litter at loader startup (pulled forward from #38)

onemen:
Part of #4 (pulled forward from #38) — tracked row: "chrome.manifest* startup sweep (branch `postv1/manifest-lifecycle-e2e`)".

## What

A killed session can leave `browser-extension-data/<id>/` littered: a 0-byte `chrome.manifest` (crash between the loader's truncate and remove) or `chrome.manifest.<uuid>` files (leftovers from uuid-named temp-manifest loader variants). Registrations are re-derived every session from what gets `autoRegister`ed, so leftovers are pure litter.

- **`core/chrome/utils/BootstrapLoader.js`** — `sweepStaleManifests()` runs at startup, before the fresh manifest is written: removes every `chrome.manifest*` file in the add-on's bed dir (missing dir = no-op; removal failures logged, never thrown).
- **`test/e2e/core/manifest-lifecycle-e2e.mjs`** — real-Firefox E2E (shipped utils used verbatim): seeds the three-file litter, then asserts across three sessions (fresh / plain restart / cache-clear restart) that chrome comes up, stays up, and the litter is gone after each clean close.
- **`test/investigate/`** — the discussion #101 investigation archive (verbatim) + the harness it produced, excluded from the format gates.

## Validation

- Rebased on current main (4 commits, no conflicts; #168's Waterfox guard untouched).
- Full E2E run against a branch-built snapshot on a portable Firefox (writable GreD): **13/13 PASS** — including the three sweep assertions that fail against a pre-sweep snapshot (9/13), proving the test actually discriminates.
- Branch CI snapshot: `dist/dev-postv1-manifest-lifecycle-e2e-aedc3b9`.


test/e2e/core/manifest-lifecycle-e2e.mjs
```diff
@@ -0,0 +1,572 @@
...
+  const savedGre = saveGreState(greDir);
+  const seedErr = (() => {
+    try {
+      return seedGre(greDir, snapshotDir);
+    } catch (err) {
+      return err.message;
+    }
+  })();
+  check(counter, !seedErr, 'seed GreD (config.js + probe)', seedErr || '');
+  if (seedErr) {
```
onemen, line 453: 🤖 AI review triage (Codebuff agent — result of the review:batch run): accepted (CodeRabbit major). The seed-failure exit skipped the finally block and could leave the Firefox install partially modified. Fixed in 510b4ac: restoreGreState(savedGre) now runs before the exit, reported as its own check.

test/e2e/core/manifest-lifecycle-e2e.mjs
```diff
@@ -0,0 +1,572 @@
...
+
+function today() {
+  return new Date().toISOString().slice(0, 10);
+}
+
+// ── Probe / lifecycle log readers ──────────────────────────────────────────
+
+function probeChunk(profileDir, byteOffset) {
+  const logPath = path.join(profileDir, 'chrome-probe.log');
+  if (!fs.existsSync(logPath)) return {size: 0, chunk: ''};
```
onemen, line 326: 🤖 AI review triage (Codebuff agent — result of the review:batch run): accepted. probeChunk returned a bare string when the probe log was absent while both callers destructure an object — the expected failure scenario would crash. Fixed in 510b4ac: returns {size: 0, chunk: ''}.

test/e2e/core/manifest-lifecycle-e2e.mjs
```diff
@@ -0,0 +1,572 @@
...
+    const canonicalOk =
+      afterS1.length === 1 &&
+      afterS1[0] === 'chrome.manifest' &&
+      canonicalContent.length > 0 &&
+      canonicalContent.includes('content testext ');
+    check(
+      counter,
+      canonicalOk,
+      'loader re-wrote its own chrome.manifest (fixed name, real content)',
+      JSON.stringify(afterS1)
```
onemen, line 517: 🤖 AI review triage (Codebuff agent — result of the review:batch run): accepted. S2/S3 now assert trailingErr === 0 like S1 — a mid-session chrome death can no longer pass the leg. Fixed in 510b4ac; full run 15/15 on a real Firefox.

Issue #130, opened 2026-09-05
E2E profile/process hygiene: deterministic repeat runs


onemen:
Part of #3 (Phase 4 — E2E matrix). Source: test-infrastructure audit absorbed into `docs/future-work.md` §6 (PR #128).

E2E runs should be repeatable on all three OSes without manual cleanup between them.

Scope:

- Kill stray installer/browser processes left by a previous run (tag spawned processes for teardown).
- `removeProfileCompatibilityIni` after profile seeding so Firefox doesn't reuse stale compatibility state.
- Deterministic seeded profiles across repeats.

Acceptance: the full matrix passes twice in a row locally and in CI with no manual intervention.

PR #248, opened 2026-09-18
test(e2e): per-leg watchdog names the wedged installer layer

onemen:
Part of #130 (hygiene follow-up) · closes out the "per-leg installer-E2E watchdogs" waiting-room item from the pre-release queue.

## Problem

The installer E2E keeps one global watchdog (`E2E_WATCHDOG_MIN`, default 10 min) that sweeps orphans and exits non-zero on a wedge — but its only message is "run exceeded 10 min". Nothing says **which layer** hung (http / test-surface / restart-scope / ui), so a CI wedge needs log archaeology to even start debugging.

## Change

- New `test/e2e/shared/legWatchdog.mjs`: `withLegWatchdog(name, fn, opts)` races the leg body against a per-leg timer. On expiry it logs `[leg-watchdog] leg '<name>' exceeded N min — killing stray installer/browser children and failing`, runs the **same** stray-process sweep as `processHygiene` (injected `run`/`platform` seams, sweep failures never mask the timeout), and fails with a leg-named error.
- All four installer layers wrapped: `http`, `test-surface`, `restart-scope`, `ui`.
- Budget: **6 min default** (`LEG_WATCHDOG_MIN` env override) ≈ 2.4× the slowest legal leg (UI layer ≤2.5 min on windows-latest); the sum of all four default-on legs stays below the 10-min global watchdog, which remains the backstop that guarantees the orphan sweep + CI log win the race against the job timeout.
- `docs/ci-inventory.md` timeout table row extended with the per-leg behavior.
- 7 unit tests (expiry naming, sweep-through-seams, sweep-throw isolation, passthrough, default fallback, unref hygiene).

## Validation

- `pnpm test`: 571 tests, 568 pass / 3 skip / 0 fail
- `pnpm lint` + `pnpm format`: clean
- Real-harness probe: forced expiry fires the leg-named log line, the real hygiene sweep, and the leg-named error
- Healthy path: full local installer E2E run shows identical results to unmodified main (9 pre-existing local-environment/snapshot failures in both cases, 0 caused by this change)

🤖 Generated with Codebuff


test/e2e/shared/legWatchdog.mjs
```diff
@@ -0,0 +1,119 @@
...
+ *   line) the leg budget so a leg timer can never let the global backstop fire
+ *   first — LEG_WATCHDOG_MIN raises the ceiling, not the deadline. marginMs
+ *   (default SWEEP_MARGIN_MS) is the cleanup room reserved from remainingMs;
+ *   log/run/platform mirror killStrayProcesses' test seams and are forwarded to
+ *   the sweep.
+ * @returns {Promise<T>} the leg body's result (falsy results preserved)
+ * @throws {Error} leg-named watchdog error once the budget expires — a leg
+ *   rejection that lands after expiry is superseded by it
+ */
+export async function withLegWatchdog(name, fn, opts = {}) {
```
onemen: 🤖 AI review triage (Codebuff agent — result of the review:batch run)

Accepted. `withLegWatchdog` awaited `fn` before the timer's reject could land, so a leg that resolved *during* expiry could mask the timeout — and the outcome was read via truthiness, so a falsy-but-real leg result could be dropped. Fixed in 29a06d0: the leg promise is settled-first-wins against the timer (expiry decided exactly once) and the outcome object is checked (`outcome.settled`), preserving falsy results. Unit tests cover both paths through the injected `run`/`platform` seams.

test/e2e/shared/legWatchdog.mjs
```diff
@@ -0,0 +1,119 @@
...
+  const log = opts.log ?? console.log;
+  let ms = opts.timeoutMs ?? resolveMs(opts.timeoutMin);
+  if (opts.remainingMs != null) {
+    const available = Math.max(1_000, opts.remainingMs - (opts.marginMs ?? SWEEP_MARGIN_MS));
```
onemen: 🤖 AI review triage (Codebuff agent — result of the review:batch run)

Accepted. A fixed per-leg budget could exceed the time *remaining* under the global run watchdog — on a slow machine a late leg's timer would fire after the global backstop, reordering the two guarantees. Fixed in 29a06d0: the leg budget is clamped to `remainingMs - margin` (default 60 s sweep margin, 1 s floor). `LEG_WATCHDOG_MIN` raises the ceiling, never past the global deadline; an override above the remaining budget is clamped and logged, not an error. Unit tests cover clamp, floor, and no-clamp paths.

test/e2e/installer/installer-e2e.mjs
```diff
@@ -1434,6 +1435,22 @@ async function run() {
...
+  // global timer above stays the backstop that sweeps orphans on a hard
+  // wedge, but it cannot say which leg hung. LEG_WATCHDOG_MIN overrides the
+  // per-leg default (6 min) for slow machines, same shape as E2E_WATCHDOG_MIN;
+  // it only raises a leg's ceiling — every leg is still clamped to the time
+  // actually left under the global deadline, so the leg timer can never let
+  // the global backstop fire first.
+  const parsedLegMin = Number.parseInt(process.env.LEG_WATCHDOG_MIN ?? '', 10);
+  const LEG_MIN = Number.isFinite(parsedLegMin) ? Math.max(1, parsedLegMin) : undefined;
+  const runStartedAt = Date.now();
+  const legOpts = {timeoutMin: LEG_MIN};
```
onemen: 🤖 AI review triage (Codebuff agent — result of the review:batch run)

Accepted (production side of the clamp finding). `legOpts` was computed once before the legs ran, so every leg inherited the *first* leg's remaining budget. Fixed in 29a06d0: the `leg()` wrapper recomputes `remainingMs` from the shared `runStartedAt` at each leg's start, so every leg is clamped against the time actually left.

onemen:
🤖 Agent housekeeping note: this PR's head branch was accidentally deleted and restored within a minute (same commit `29a06d0`, force-free — no history changed). GitHub auto-closed the PR on the deletion; it was reopened immediately. CI may re-run; no action needed from reviewers.

PR #249, opened 2026-09-18
docs: record cross-channel validation of the manifest sweep (#245)

onemen:
Follow-up to #245 (chrome.manifest\* startup sweep): validate it against the browser channels CI does not run, per the operator's request.

## Local validation — all PASS

Fresh `main` snapshot (contains #245), `test/e2e/core/manifest-lifecycle-e2e.mjs --headless`, three sessions each (fresh profile / plain restart / cache-clear restart):

| Browser | Version | Result |
|---|---|---|
| Firefox Nightly (local copy of `C:\Program Files\Firefox Nightly` for a writable GreD) | 158.0a1 | **15/15 PASS** |
| Firefox ESR (`esr_previous`) | 140.16.0esr | **15/15 PASS** |
| Firefox ESR 128 (`D:\My Documents\...`) | 128.14.0esr | **15/15 PASS** |

Sweep-specific assertions green on all three: seeded manifest litter removed at loader startup, fresh manifest is the only file left, chrome live across all three sessions, no trailing probe ERR.

## Why there is no CI change

The requested "skip ESR 128 in CI" turned out to be a no-op: `grep -rn "esr" .github/` returns nothing — **CI has no ESR legs**. The updater/installer hard gate is Firefox stable + Dev Edition + Nightly (first-party channels, #35); ESR coverage rides on the shared chrome lifecycle. So this PR only records the validation evidence in `docs/future-work.md` (the home of #130-adjacent follow-ups) with the re-run rule: re-run this suite on an ESR build after any loader lifecycle change.

🤖 Generated with Codebuff
Co-Authored-By: Codebuff <noreply@codebuff.com>

PR #250, opened 2026-09-18
ci: run the core lifecycle E2E in CI; script-coverage contract; cleanup

onemen:
Corrective PR after the #245 review findings: the sweep shipped with a test that nothing ever ran, plus repo clutter. Fixes the gap and adds the guard so it cannot recur.

## 1. The core lifecycle E2E now runs in CI (the actual fix)

New **`core lifecycle E2E · <os>`** job in `e2e.yml` — **REQUIRED**, gated on the `core` path filter, all 3 OSes, portable Firefox (writable GreD), running `test:e2e:legacy` (the manifest-lifecycle harness) via the same command CI uses:

- Registers a legacy bootstrap extension's chrome across **3 real Firefox sessions** (fresh profile / plain restart / cache-clear restart) — the #101 regression surface where `BootstrapLoader.js` + `userChrome.js` actually break.
- Asserts chrome stays live every session, no trailing errors, and #245's startup sweep leaves **no manifest litter**.

From this PR on, **every PR that touches `core/**` proves the legacy chrome lifecycle on CI's Firefox**, on all three OSes. The updated job lands in the e2e-gate's `required:` + `applicability:` + `skip-guard:` blocks; the gate-coverage contract is updated to match.

## 2. Script-coverage contract — the guard

`tools/check-gate-coverage.mjs` (already running in CI as `pnpm check:gates`) gains `checkTestScriptCoverage()`: every `test:*` script in package.json must be referenced by a workflow, a script chain, or the new `MANUAL_TEST_SCRIPTS` allowlist (with a documented reason — currently only the local convenience wrappers + `test:skills`). The checker **demonstrably catches the #245 failure**: running it on the pre-PR tree reports `test:e2e:legacy` as unreferenced. Also fixes prettier-wrapped continuation lines in verify-gate `with:` blocks silently losing their tail during parsing. 4 new unit tests.

## 3. Cleanup

- **`test/investigate/` deleted** — the discussion-#101 archive and the superseded investigation harness were working artifacts, not repo content.
- **`test/e2e/updater/puppeteer/` deleted** — a personal debug probe (hardcoded `C:\Users\Hadar\...` profile path, its own README); not a test, not wired anywhere.
- **#245's gate exclusions reverted** — the `verbatim-archives` eslint block, the `.prettierignore` entry, and the markdownlint exclusion existed only to tolerate `test/investigate/`.

## Docs

- `docs/ci-inventory.md`: the core-lifecycle row + trigger/required semantics.
- `docs/future-work.md`: the manual cross-channel record preserved as evidence (Nightly 158.0a1 / ESR 140.16.0esr / ESR 128.14.0esr — all 15/15, 2026-09-18) and the tracked follow-up: channel-keyed `firefox-esr`/`firefox-esr-next` watchdog rows + advisory legs (auto-rotating with Mozilla's ESR transitions: 153 becomes the served ESR on 2026-09-29, 164 on 2027-01-26).

## Validation

- `pnpm test`: 568 tests, 565 pass / 3 skip / **0 fail** (23/23 in the gate-coverage suite)
- `pnpm lint`, `pnpm format`: clean; `pnpm check:gates`: contracts hold (14 + 4 jobs)
- Local harness run via the CI command shape: **15/15 PASS** against the portable Firefox

Closes #249 (superseded — that PR documented manual validation as prose instead of fixing the CI gap; its evidence table lives on here).

🤖 Generated with Codebuff


config/.markdownlint-cli2.jsonc
```diff
@@ -47,11 +47,6 @@
     // END managed
     "**/node_modules/**",
     "**/dist/**",
-    // Verbatim archive (test/investigate/README.md): discussion threads
-    // preserved byte-for-byte — MD034 etc. would force edits that break the
-    // verbatim guarantee (same rationale as the verbatim-archives block in
-    // config/eslint.config.js and the .prettierignore entry).
-    "**/test/investigate/*.md",
     // Local drafts (gitignored .local files/dirs) — same exclusion
     // .prettierignore and eslint.config.js use (AGENTS.md: they are not
```
onemen, line 51: 🤖 AI review triage (Codebuff agent — result of the local ai-review run on this PR)

**Finding (triaged: WRONG in premise, accepted as the trigger for a doc note).** The reviewer warns that removing `**/test/investigate/*.md` from markdownlint ignores breaks the byte-for-byte preservation guarantee of the archive. Premise no longer holds: this PR *deletes* `test/investigate/` entirely, so there are no archive files left to lint — the exclusion had nothing left to exclude.

The underlying intent (a note explaining *why* the entry disappears) is fair; addressed by documenting the deletion rationale in the commit message and PR body rather than keeping a dead config entry. No code change needed.

tools/check-gate-coverage.mjs
```diff
@@ -411,12 +427,133 @@ const CONTRACTS = [
...
+      if (texts.some(t => pattern.test(t))) {
+        reachable.add(name);
+        texts.push(stripYamlComments(String(allScripts[name])));
+        grew = true;
+      }
+    }
+  }
+
+  for (const script of testScripts) {
+    if (MANUAL_TEST_SCRIPTS.has(script)) continue;
```
onemen: 🤖 AI review triage (Codebuff agent — result of the review:batch run)

Accepted. The first cut treated every script's command text as a surface, so an orphaned wrapper could "cover" the script it invokes — and `includes()` let `test:e2e` match inside `test:e2e:legacy`. Fixed in 380e196: a script's command text joins the surfaces only once the script itself is reachable from an automation root (a real closure), names match exactly, and YAML comments are stripped. Negative tests added for the orphan wrapper, prefixed-name, and commented-invocation cases.

PR #255, opened 2026-09-19
fix: add resource limit in main.c (CWE-770)

anupamme:
The installer_set_uploaded_zip and installer_set_uploaded_ui_zip functions accept arbitrary-sized uploads without enforcing an upper bound. While basic validation exists (null check, zero-length check, malloc success check), an attacker with a valid session token could upload extremely large files causing memory exhaustion and denial of service. The affected code is `installer/src/main.c:364`. This change is the fix I would apply.

Reference: [CWE-770](https://cwe.mitre.org/data/definitions/770.html)

## What changed
- `installer/src/main.c`

## Verification
No automated check could be run against this repository, so this change is unverified beyond review. Please treat it as a suggestion.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


onemen:
🤖 Codebuff agent — deferral decision (per the operator, 2026-09-20):

This PR is **deferred to post-release**. It changes `installer/src/main.c`, so merging it now would shift the `installer_win.exe` bytes and invalidate the WDSI-cleared build the v1.0 runbook ships (the three false-positive submissions and the SmartScreen guidance are bound to those bytes).

Assessment: the hardening is reasonable, but the exposure is loopback-only, token-gated (128-bit session token), and single-user — memory exhaustion via an authenticated local upload is not a realistic pre-1.0 threat. This will be the first post-release security item; the next installer release (and any future WDSI submission) carries it.

PR #264, opened 2026-09-20
ci: pin ubuntu-24.04 runners + advisory ubuntu-26.04 canary + runner annotations watchdog

onemen:
## Why

`ubuntu-latest` flips to Ubuntu 26.04 starting **2026-10-19** (completing by 2026-11-19) with a **gradual, nondeterministic rollout** — during that window identical jobs land on either image (actions/runner-images#14748; the notice GitHub attaches to every `ubuntu-latest` job, seen on run 35505810014). The real risk here is the **shipped Linux binaries**: `installer_linux` / `installer_linux_aarch64` are compiled with gcc installed from apt on the runner image, so an image flip changes artifact bytes silently mid-rollout — exactly what the pinned-MSYS2 philosophy prevents on Windows, and the Linux side had no equivalent pin.

## What

1. **Pin** — every `runs-on: ubuntu-latest` across all 7 workflows → `ubuntu-24.04` (deterministic through the rollout window). Includes the E2E display names and `record-validated-versions`' `UPDATER_LEG_OSES` (the validated-versions record matches leg OS names).
2. **Canary** — `CI / publish gate canary — ubuntu-26.04`: the Linux build job's steps cloned verbatim onto `ubuntu-26.04`, **advisory** per the gate-contract classification. A red canary means "fix the build on 26.04 before moving the pins" — never blocks a merge/publish. Upgrade later = a deliberate one-line pin bump.
3. **Runner watchdog** (`tools/runner-watchdog.mjs` + `runner-watchdog.yml`, the skills-watchdog pattern): weekly Monday 20:00 UTC scan (staggered off URL 06:00 / skills 14:00) of
   - NOTICE/warning annotations on the newest run per workflow (`/check-runs/{id}/annotations`), filtered to migration/deprecation phrasing;
   - new `Announcement` issues in `actions/runner-images` (8-day lookback).
   Findings land in **one rolling `[runner-watchdog] CI runner deprecations` tracking issue** (deduped by title, auto-closed when a later run is all-clear). PR leg (stateless, `::warning::`, always green) triggers only when the watchdog itself changes. Runner *labels* are not covered by Dependabot — this is their monitor.
4. **Docs** — `docs/ci-inventory.md` synced: new canary row, watchdog rows, timeout note, leg-name renames.

Dependabot already covers the `github-actions` ecosystem (pinned SHAs) — no config change needed.

## Validation

- `pnpm lint` (incl. gate-coverage contract on the new canary job + MD056 table integrity) — green
- `pnpm format` — green
- `pnpm test` — 642 tests, 0 fail (7 new watchdog tests: classification, dedup, announcement filtering, issue-body rendering, per-job API-failure tolerance)
- Pre-push gate — green

Refs actions/runner-images#14748

🤖 Generated with Codebuff


onemen (commented):
🤖 AI review triage (Codebuff agent — result of the operator-initiated `pnpm review:batch` run on this PR)

Batch review returned 4 findings (3 major, 1 minor). Triage: **4 accepted, 4 fixed** in b1b43d7 — all verified against the code before applying:

1. **Missing `actions: read` / `checks: read` scopes** (major) — **right**. Both jobs passed only `contents: read` (+`issues: write`), so the workflow-run/job and check-run-annotation queries would 403; the advisory top-level catch would have hidden it as a green no-op. Both scopes added to `check` and `pr-check`.
2. **Absolute URLs into `makeFetchJson`** (major) — **right**, and the best catch of the run: the helper prepends `API_ROOT`, so the announcement search and the tracking-issue lookup built `https://api.github.comhttps://api.github.com/…`. The announcement scan never worked, and the lookup failure landed outside its recovery path. Both call sites now pass relative paths.
3. **All-clear after an incomplete scan** (major) — **right**. Swallowed lookup failures could blank the findings and trigger the auto-close of a live tracking issue. `scanRunAnnotations` now reports `incomplete`; an incomplete scan keeps any open tracking issue open and says so in the run log / `::warning::`.
4. **Vacuous assertion** (minor) — **right**. `assert.ok(calls.some(... === false))` passed whenever *any* call lacked the path; flipped to a hard `assert.equal(..., false)` and added a clean-scan (`incomplete === false`) test.

Validation after the fix: full suite 643/0 fail (8 watchdog tests), `pnpm lint` + `pnpm format` green.

PR #267, opened 2026-09-20
watchdog: triage view in the runner tracking issue (#265 UX)

onemen:
## Why

[#265](https://github.com/onemen/firefox-scripts/issues/265) works, but it's a firehose: every finding looks equally open, nothing distinguishes what we can act on from GitHub-managed noise, and there is no record of what's already been handled. (Operator feedback after the first live run.)

## What

The tracking-issue body becomes a **self-maintaining triage view**:

1. **Ours vs. External split** — findings from workflow files in this repo are actionable ("Ours"); findings from GitHub-managed runs (`pages build and deployment`, Dependabot's own checks) are informational ("External — nothing in a commit here can change them"). Classification is by the run's `path` (`.github/workflows/<file>`), not the workflow name: a Dependabot-triggered run of *our* `ci.yml` reports the PR title as its name (observed live in #265) but still carries our path. Ambiguity (no path, missing tree) classifies as external — never silently dismissed.
2. **First-seen / last-seen dates** on every row, persisted across rewrites.
3. **Human triage via checkboxes** — tick an item `[x]` to mark it handled. The tick survives every weekly rewrite: `parseLedger` harvests rendered checkboxes (the human's edit surface) and merges them over the machine ledger comment at the end of the body. Ticked items move to a collapsed "Handled" provenance section.
4. **Auto-clear** — a tick whose finding stops firing disappears from the live sections (stale state never resurrects); the next all-clear scan closes the issue as before.

`docs/ci-inventory.md` synced per the must-sync rule.

## Validation

- `pnpm lint` (incl. MD056 after table reflow), `pnpm format`, `pnpm test` — 647 tests / 0 fail (12 watchdog tests: classification, ledger round-trip, announcement-tick harvest, stale-tick auto-clear, all-clear wording)
- Pre-push gate green

🤖 Generated with Codebuff


onemen (commented):
🤖 AI review triage (Codebuff agent — result of the ADR 0020 `pnpm review:local` run on this PR)

1 finding, assessed **wrong** — not applied:

- **"stale tick auto-clears test contradicts the Handled section"** — the reviewer read the Handled section as preserving ticks forever. It renders only for items present in the **current** scan (handledEverything is built from findingRows/announcementRows, which exist only for findings that fired this run). When a finding stops firing, no row exists, so nothing renders with a tick — the old tick survives only inside the ledger blob, deliberately (the test comment says exactly this: the issue auto-closes on all-clear anyway, so keeping a zombie Handled row for a finding that no longer fires would contradict the section's purpose). The assertion `doesNotMatch /- \[x\]/` matches the implemented and intended behavior; the full suite passes against it.

Provider-side notes, not findings: the `tools/runner-watchdog.mjs` review was skipped by a transient provider outage and the `ci-inventory.md` reply was invalid JSON (skipped) — the two files were exercised by the batch review process earlier today and by the full gate suite.

onemen (commented):
🤖 AI review triage (Codebuff agent — result of the operator-initiated `pnpm review:batch` run on this PR)

Batch review returned 3 findings (3 major). Triage: **3 accepted, 3 fixed** in c870cb8 — all verified against the code first:

1. **`buildIssueBody` passed `f.workflows` (names) to `isOursFinding`, which expects run paths** (major) — **right**, the best catch: every repo finding would have classified External, making the Ours section permanently empty. The earlier local-review run missed it because my section assertions were section-agnostic. Fixed, and the test now asserts the e2e.yml finding renders *inside* the Ours section.
2. **Unticked checkboxes could not clear handled state** (major) — **right**: `parseLedger` only harvested `[x]`, so a maintainer's untick was silently re-ticked on the next rewrite. Both checkbox states are now authoritative; a test pins that an untick stays unticked and the Handled section disappears.
3. **A failed body-fetch would erase all triage state** (major) — **right**: `previousBody` stayed null and the subsequent PATCH would rewrite the body with fresh dates and no ticks. The run now leaves the issue unchanged when the body is unreadable, and retries next week.

Validation after the fix: 648 tests / 0 fail (13 watchdog tests), lint + format green, pre-push gate green.

PR #266, opened 2026-09-20
docs: apply queue B — all docs fixes from the 2026-09-18 audit §7

onemen:
Part of the pre-release cleanup (queue B, ~3 h estimate). Applies **every docs-only item** from the 2026-09-18 working-tree audit (local draft, not committed) §7 in one PR. **Docs-only: zero bytes of `installer_win.exe` change.**

## What's in

| Item | Files | Fix |
|---|---|---|
| **P0 D1/D2** | `docs/security.md`, `docs/restart-ui-tab.md` | session token `16-hex` → `32-hex` (128-bit since #209) |
| **P1 D3/D4/D5** | `docs/DEVELOPING.md` | generated-files section: "Three" → four files, `generatedRegistry.mjs` named as the owner, `updater.css` built by `createZip.mjs` (not `uploadToPages.mjs`), missing space |
| **P1 W1/D11** | `docs/DEVELOPING.md` CI section | documents the load-bearing `$/` local-action form (19 uses, 4 workflows) — the "cleanup" that breaks runner-only jobs is now explicitly forbidden in prose |
| **P1 (2.3)** | `docs/decisions/0032-…` + link | ADR 0032 records the pinned MSYS2 toolchain (the decision that determines shipped bytes; context #157) |
| **P2 C7/D12** | ADR 0010 | glued dated notes split into separate paragraphs |
| **P2 C8** | ADR 0010 | strcmp-not-constant-time threat-model note recorded |
| **P2 D7** | `docs/roadmap.md` | §6 pointer now names the resolved gaps + the one open item (#33) |
| **P2 D13** | `docs/future-work.md` | §2.2 heading names the decided tool (puppeteer-core, ADR 0015) |
| **P2 D8** | `CONTRIBUTING.md` | verification block: 7-stage `pnpm lint`, `test:hash` snapshot precondition, `--` flag form |
| **P2 D10** | `docs/decisions/0033-…` + `.github/dependabot.yml` | ADR 0033 records the dependency-update policy; the mis-attribution to ADR 0017 is fixed |
| **§5 coverage** | `docs/security.md` | API route list cites `apiRoutes.mjs` as the CI-enforced source of truth (unit contract test + smoke test), instead of a drifting prose copy |
| **S1/S4/S5** | skills + `AGENTS.md` | `test:hash` precondition in change-workflow + publishing validate tables, validate-table source-of-truth note, vendored-skills host-dependence sentence |
[...]

PR #268, opened 2026-09-20
test: share the snapshot-provenance guard (T1); anchor the process-sweep matcher (T5)

onemen:
Two audit items from the 2026-09-18 working-tree audit (local draft, not committed) that harden release day itself. **Test-only: zero bytes of `installer_win.exe` change.**

## T1 — `test_hash.mjs` validates a stale snapshot (P1)

`test_hash.mjs` ran whatever binary sat in the newest `dist/` snapshot with no provenance check; its sibling `test_self_update.mjs` refused stale ones (2026-09-15 fix). Parity "proven" against an old build proves nothing about `main` — and `test:hash` is a release-day gate.

**Fix:** extract the guard into shared `installer/test/snapshotProvenance.mjs` (`computeInstallerSourceHash` + `requireFreshSnapshot` with an injectable `exit` seam), call it from **both** suites, unit-test the drift / corrupt-manifest / missing-hash paths.

**Verified live:** the newest snapshot on this machine (`dev-main-450468f`, built from `d850ecf…`) is refused against current sources (`f5e1a3d…`) — the exact mismatch the audit cited.

## T5 — process-sweep regex over-matches (P2)

The sweep matched `installer_win` as a bare **substring**, so any shell whose command line contained the literal was killed — verified live in the 2026-09-15 follow-up (a calling shell died, exit 255). Release day means many commands naming the installer.

**Fix:** the matcher now requires the binary in the **argv[0] position** (optionally quoted path): the harness always *spawns* the binary, while `ls dist/installer_win.exe` / `grep installer_win …` have it as an argument and are no longer touched. One matcher, three dialects — JS regex, PowerShell `-match` (single-quoted literal, no doubling needed), POSIX ERE for `pkill -f` (no `(?:…)`, `[[:space:]]` not `\s`) — all three exported and **validated against their real engines** in unit tests: actual `powershell.exe` on Windows hosts, actual `grep -E` everywhere (skips where unavailable).

## Validation

- `pnpm lint` — 7/7 stages, 0 findings
- `pnpm format` — clean
- `pnpm test` — **649 pass / 0 fail** (was 632: +5 provenance, +3 matcher/T5, +2 engine-backed, +9 reshaped)

Fixes the two release-adjacent items of the 2026-09-18 audit; C1/C2/C4/C5/C6, manifest.c, the testing-boundary ADR and provenance attestation remain deliberately post-release.

🤖 Generated with Codebuff


test/unit/installer/snapshotProvenance.test.mjs
```diff
@@ -0,0 +1,159 @@
...
+import {test} from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+const REPO_ROOT = fileURLToPath(new URL('../../..', import.meta.url));
+
+// publishCommon/paths.mjs (in the import chain below) demands a --mode at
```
onemen, line 20: 🤖 AI review triage (Codebuff agent — result of the local `pnpm review:local` run on PR #268)

**Accepted.** The `--mode=prod` argv push is the established house pattern (the zip/publish unit tests and both binary-in-the-loop suites do the same), so the behavior is unchanged — but this file carried no explanatory comment, which is exactly how a subtle pattern becomes a trap. Added the same comment those files use in 38a891e (import-time --mode requirement + `node --test` per-file process isolation keeping it scoped).

test/unit/e2e/processHygiene.test.mjs
```diff
@@ -15,8 +15,96 @@ const REPO_ROOT = fileURLToPath(new URL('../../..', import.meta.url));
...
+    assert.equal(out.length, cases.length, 'one verdict per case');
+    cases.forEach((c, i) => {
+      assert.equal(out[i], c.want, `PowerShell verdict for ${JSON.stringify(c.cmd)}`);
+    });
+  });
+}
+
+test('INSTALLER_ARGV0_ERE: real grep -E (POSIX ERE) agrees with isE2eProcess', async t => {
+  const {spawnSync, execFileSync} = await import('node:child_process');
+  // Windows does not ship grep (the pkill branch of the sweep is dead code
```
onemen, line 84: 🤖 AI review triage (Codebuff agent — result of the CodeRabbit review:batch run over PRs #266 + #268)

**Accepted (Major).** A standard Windows environment provides no grep, so the check would fail with ENOENT right after the PowerShell check passed. Fixed in 62c5151: probe `grep --version` and `t.skip()` when unavailable — capability-based rather than platform-based, so Git Bash / WSL hosts and Linux CI still run the real check.

PR #271, opened 2026-09-20
fix(updater-ui): broken helper checksum verification + CSP-blocked inline style

onemen:
Two `updater-ui.zip` bugs found while manually testing the `upload:local` snapshot against an admin-protected Firefox install (`C:\Program Files\Firefox Nightly`). **No installer byte changes** — both fixes ship in the updater tab package only.

## 1. Helper checksum verification always failed (release-blocker for elevated installs)

`fileSha256Hex` used `hasher.finish(false)` — a binary *string* — and spread it into UTF-16 characters, calling `toString(16)` on each. Bytes ≥ 0x80 came out verbatim (`got 0fØ•0ØÆ0Ø6\/…` in the console), so the digest never matched the published sidecar and `ensureHelper` refused every downloaded helper. The elevated `config.js` / `defaults/pref/config-prefs.js` install aborted before any copy, and the helper has never run in any browser install whose directory is not user-writable. Broken since #174.

Why CI never caught it: the updater/installer E2E harnesses deliberately run against user-writable portable installs (direct-copy path; both harnesses *fail fast* if GreD is not writable), so the helper is never exercised in CI. The admin-protected helper path is covered by this fix's source-invariant test now; a CI leg for it is queued as post-release work.

Fix: use the portable conversion the engine already uses for the utils/fx-folder package hashes (`finish(true)` → `atob` → `charCodeAt` per byte) + a unit test pinning the pattern so `finish(false)` cannot reappear.

## 2. CSP-blocked inline style on the progress-bar fill

#228's CSP meta (`style-src chrome://firefox-scripts`, no `'unsafe-inline'`) made Firefox log a `style-src-attr` violation for the inline `style="width: 0%"` on the progress-bar fill (present since the initial native-installer commit; harmless before the CSP). The attribute was redundant: the stylesheet already defaults `.card-progress-bar-fill { width: 0% }` and the runtime updates go through `el.style.width` (CSSOM — CSP-allowed).

Fix: remove the attribute + a `cspMeta` assertion pinning "no inline style attributes" on the updater tab (scoped there — the installer page's two inline styles are embedded in `installer_win.exe` and wait for the post-release CSP cleanup).

## Validation

- `pnpm test` — 655 pass / 0 fail (two new pins; suite 658)
- `pnpm lint`, `pnpm format` — clean
- Manually verified pre-fix failure (mojibake checksum) and post-fix conversion on real PE bytes

Part of #38 (post-#228 CSP hardening follow-up).


onemen:
🤖 Found while manually testing the `upload:local` snapshot (helper checksum rejected every download with mojibake hex). Root cause + fix in `702b308`..HEAD: `fileSha256Hex` spread a binary string (finish(false)) instead of converting per-byte — bytes ≥ 0x80 came out verbatim, so verification could never pass. Elevated installs (admin-protected browser dirs) were broken since #174; user-writable installs never run the helper, which is why CI never caught it. Fixed with the engine's portable conversion + regression pin; raised CI priority accordingly — the installer E2E legs do exercise the helper path on admin-protected dirs.

onemen:
🤖 Correction to my earlier comment: I claimed the installer E2E legs exercise the helper path — **wrong**. Both the updater and installer E2E harnesses deliberately use user-writable portable installs (they fail fast with "GreD not writable" otherwise), so the helper never runs in CI. That's exactly why the mojibake checksum survived #174 unnoticed. A CI leg covering the admin-protected helper path needs deliberate ACL-denied GreD on a Windows runner — queued as a separate post-release PR.

test/e2e/updater/updater-e2e.mjs
```diff
@@ -92,12 +99,23 @@ try {
...
+        let line;
+        try {
+          const se = aMessage.QueryInterface(Ci.nsIScriptError);
+          const errFlag = Ci.nsIScriptError.errorFlag || 1;
+          const warnFlag = Ci.nsIScriptError.warningFlag || 2;
+          const level =
+            se.flags & errFlag ? 'error' : se.flags & warnFlag ? 'warn' : 'info';
+          const src = se.sourceName ? ' [' + se.sourceName + ':' + (se.lineNumber || 0) + ']' : '';
+          line = level + src + ' ' + se.errorMessage;
+        } catch {
```
onemen, line 119: 🤖 AI review triage (CodeRabbit review:batch run, assessed by the Codebuff agent) — accepted.

Testing `errorFlag || 1` first is version-fragile: the nsIScriptError flag constants moved between Firefox versions (a zero-valued/removed `errorFlag` makes the `|| 1` fallback test a bit that is not "error" any more, misclassifying info as error). Classify info → warn → everything-else-as-error using `infoFlag`/`warningFlag` with new-scheme fallbacks; unknown flags default to error so the console net can only over-report, never under-report.

Issue #33, opened 2026-08-24
Publish pipeline automation: build-and-upload.yml, deterministic-publish check, helper sha256 trust


onemen:
From docs/future-work.md §3:
- Re-add a cross-platform build-and-upload workflow (each OS builds its platform, one job publishes the staged set).
- Deterministic publish output check: run upload:local --mode=prod twice, fail if the snapshot (hashes, zips, manifest files lists) differs.
- [x] Publish helper_<platform>.sha256 assets and assert the downloaded helper matches before execution — merged in #174 (9a945a4)
- [x] Date-stamped component releases (scripts-<date> / installer-<date>) alongside `latest` — see issue #72, ADR 0019 — merged in #176, refined in #189 (Latest-badge re-pin with make_latest; no prerelease trick)

- [x] Staging-target guard: when env overrides redirect publish targets away from prod (`REPO_OWNER`/`ZIP_PAGES_BRANCH`/`RELEASE_NAME`/`HASHES_URL`/…), show a STAGING banner and fail/warn loudly instead of silently publishing to an unintended target (env-over-conf precedence already exists in `tools/publish/paths.js`)
- [x] Document the staging keys in `.env-example` (same key set, with prod-safety warnings) — merged in #164
- [ ] `msys2/setup-msys2` release caching on the Windows legs (`cache: true`) — trade freshness for minutes per job, mirroring the cached `-fanalyzer` leg (#105/#106)
- [ ] Nightly publish trigger (`schedule: cron`) — only after the staging guard is green and a publish-path E2E run has passed

PR #272, opened 2026-09-20
fix: add path validation in helper_mac.c (CWE-22)

anupamme:
The helper binary receives source and destination file paths as command-line arguments and performs file copy operations without validating that the destination path is within an expected directory. Paths are used directly in shell commands (mkdir -p, cp) and copyfile() calls without explicit path traversal guards. This was found by static analysis at `installer/src/helper/helper_mac.c:1`. I have not demonstrated an exploit against your deployment, so please judge it against your own threat model.

Reference: [CWE-22](https://cwe.mitre.org/data/definitions/22.html)

## What changed
- `installer/src/helper/helper_mac.c`

## Verification
No automated check could be run against this repository, so this change is unverified beyond review. Please treat it as a suggestion.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


onemen:
Thanks for the report and the patch — good instinct to look at the helper, since it's the one artifact that runs outside the browser sandbox. We reviewed it against the actual threat model and are closing this rather than merging, for two reasons:

1. Threat model: there is no crossed privilege boundary. The helper is a short-lived binary invoked only by our own updater tab, with argv built by our code from checksum-verified package state; it's downloaded over HTTPS from a fixed URL and verified against a published .sha256 sidecar before execution (issue #33; regression fixed in #271). For the traversal check to matter, an attacker must already control the updater's argument construction — i.e. they can already run arbitrary code as the user, and elevation via our helper grants them nothing they couldn't get by requesting it directly. We'll still take the guard as defense-in-depth, but not in this form (point 2).

2. The patch itself has an off-by-one. has_path_traversal tests path[i + 1] and path[i + 2] before confirming those indices are in bounds; for a .. at the end of the string, path[i + 2] reads past the NUL terminator. On typical allocators this stays inside the buffer's slack, but it's precisely the class of bounds bug the check exists to prevent — it would not pass our gcc -fanalyzer gate. A corrected, bounds-checked version of the guard is planned in #274 (with an explicit decision on source vs. destination paths, tests, and the full threat-model write-up).

Closing as not-planned rather than merging: the concern is noted and carried forward with the right scope in #274. Thanks again for the report.

anupamme:
Thanks for the detailed review. I agree with the threat-model distinction and with the off-by-one issue in my has_path_traversal() implementation.

I’ll treat the original finding as defence-in-depth rather than claim it as a demonstrated privilege-boundary vulnerability. The bounds-safe traversal check, scope of source vs. destination validation, and accompanying tests make sense to address together as part of #274.

PR #277, opened 2026-09-21
fix(installer): CSP connect-src allowlist for the remote ingest hosts

onemen:
## The bug (found in the 2026-09-21 dev-channel manual test)

PR #228 (audit P3) added `connect-src 'self'` to the installer page — correct for `/api/*`, but it also **forbade every remote fetch the tab's JS must make**: the architecture (ADR 0010) has the C installer do zero network I/O while the tab fetches the package URLs `/api/package-urls` hands over and POSTs the bytes to the local server. On any CI build (prod or dev), the tab could no longer reach the package server at all:

- the console showed `CSP: ... blocked ... connect-src 'self'` for every zip,
- "Prefer to install manually?" links failed silently (console-only errors),
- earlier runs missed it because `--local` snapshots bake `http://localhost:8777` URLs — same-origin, so allowed.

The updater tab is unaffected (privileged `Downloads` API, `connect-src 'none'` is correct there) — which is why updater E2E legs stayed green.

## The fix

| Change | Detail |
|---|---|
| CSP allowlist | `connect-src 'self'` + exactly the hosts `/api/package-urls` can emit: `onemen.github.io`, `api.github.com`, `cdn.jsdelivr.net`, `raw.githubusercontent.com` (dev helper), `hg.mozilla.org` |
| **Contract test** | `cspMeta.test.mjs` derives the host set from the **generated config** for prod AND dev and fails if the CSP lags — it caught the `raw.githubusercontent.com` omission *inside this very PR* |
| Manual-download names | Save name follows the URL (`fx-folder-dev.zip` on dev, not a hardcoded `fx-folder.zip`) |
| Manual-download UX | Failures show an inline error next to the links (were console-only) |
| E2E UI-12 | The installer E2E UI leg now asserts the tab's ingest settles **without the network-error banner** — a blocked remote fetch fails CI instead of shipping |

## Byte impact

Changes embedded HTML/JS/CSS → **new installer bytes** (coordinated pre-release re-roll, per the runbook plan; VT scan + hash re-pin happen on the train afterwards).

## Validation

- [x] `pnpm test` — 670 tests, 0 failures (incl. the new contract test)
- [x] `pnpm lint` + `pnpm format` clean; `resources.h` regenerated
- [x] E2E UI-12 added (runs in CI's installer E2E UI layer)
- [ ] Post-merge: rebuild + VT scan on the re-roll train; manual Nightly re-test

Part of #4 (release-day runbook preconditions).

🤖 Generated with Codebuff


installer/web/script/50-init.js
```diff
@@ -91,6 +91,18 @@
...
+     * the console) can observe the remote-fetch pipeline itself instead of
+     * inferring it from rendered cards — UI-05 stayed green through the
+     * 2026-09-21 CSP regression precisely because cards render before the
+     * remote fetches. Lifecycle: pending → complete (zips ingested) | failed
+     * (zips unreachable — banner shown) | blocked (no URLs from the server;
+     * banner shown). */
+    function setIngestState(state) {
+      document.body.setAttribute('data-ingest', state);
+    }
+    setIngestState('pending');
```
onemen, line 104: 🤖 AI review triage (Codebuff agent — result of the review:batch run 2026-09-21)

**Accepted — Major (review:batch, Functional Correctness).** The original UI-12 condition (no network-error banner + cards/empty-state present) was already true right after UI-05: cards render before `ingestRemoteData` runs, and the banner starts hidden — so UI-12 could pass while `connect-src` blocked every remote fetch, i.e. it did not guard the regression it was added for.

**Fix (commit 0ff48d7):** the tab now publishes an explicit ingest state machine on `<body data-ingest>` — `pending` → `complete` (package zips ingested) | `failed` (zip fetch failed, banner shown) | `blocked` (no URLs from /api/package-urls) — and UI-12 waits for a terminal state and passes **only** on `complete`, additionally asserting the banner is hidden.

installer/web/script/40-install.js
```diff
@@ -560,13 +560,37 @@
     link.addEventListener('click', function (e) {
       e.preventDefault();
       if (link.getAttribute('href') && link.getAttribute('href') !== '#') {
+        clearDownloadError();
```
onemen, line 563: 🤖 AI review triage (Codebuff agent — result of the review:batch run 2026-09-21)

**Accepted — Minor (review:batch, Functional Correctness).** `showDownloadError` had no hide path: a failed manual download kept its message visible even after a later download succeeded.

**Fix (commit 0ff48d7):** `clearDownloadError()` empties the message and re-hides the element before every new download attempt (called at the top of the click handler in `wireDownloadLink`), so a stale error never outlives the retry that supersedes it.

onemen (commented):
🤖 AI review triage (Codebuff agent — result of the review:batch run 2026-09-21)

review:batch (2026-09-21) reported 1 Major + 1 Minor on this PR. Both triaged as real and accepted; both fixed in 0ff48d7. Threads below are line-anchored to the fixed code.

installer/web/script/50-init.js
onemen: Fixed in 0ff48d7: explicit <body data-ingest> state machine (pending → complete | failed | blocked); UI-12 now waits for a terminal state and passes only on "complete".

installer/web/script/40-install.js
onemen: Fixed in 0ff48d7: clearDownloadError() empties and hides the message before every new download attempt.

